PR #2, opened 2024-10-24
check to see if None

codegod100:


codegod100:
looks like there is another place in the code that does an explicit panic if the value in None so maybe this isn't the best route. See #6

PR #8, opened 2024-10-24
check to see if None

codegod100:
Not sure the best way to not have panics when I print my models. Not sure if I'm missing something


carllerche (approved):
Thanks!

PR #9, opened 2024-10-24
open with file

codegod100:
I'm returning a Result value because I'm not a fan of panics. Not sure if you wanted to keep the unwraps or not.


codegod100:
Sorry if the open/close PRs were confusing, I was trying to create a flow where I could create discrete pull requests, while working on all of them at once on localhost. I think I got it figured out

carllerche (approved):
Thanks!

PR #11, opened 2024-10-24
Sqlite open function to allow connection to a sqlite file

alaingilbert:
This pull request allow you to actually open a connection to a sqlite file.

In the current state, you can only open a in-memory database.

Example:

```rs
use toasty_sqlite::Sqlite;
// ...
let driver = Sqlite::open("my_database.db").unwrap();
```


codegod100:
See also #9. At least people want to see the same features

carllerche:
I merged #9 already, but I think `Sqlite::open` probably is a better name. Feel free to open a new PR that renames the method to `open`.

Issue #3, opened 2024-10-24
use tracing e.g. `debug!` instead of `println!`


codegod100:
Can control log level and better instrumentation tooling

carllerche:
Closed by #16

PR #12, opened 2024-10-24
use tracing crate, debug! macro, instead of println/dbg

alaingilbert:
To avoid spamming stdout of projects that are using the orm library


codegod100:
Thank you See #3

carllerche:
Thanks for the PR! I meant to delete all of this before opening the git repo. Woops! I merged #16 which just removes all of the noise. Eventually, toasty will probably depend on tracing, but that should be more deliberate.

PR #19, opened 2024-10-24
rename new_with_file -> open

alaingilbert:
as discussed here https://github.com/tokio-rs/toasty/pull/11#issuecomment-2436197118


carllerche (approved):
Thanks.

PR #25, opened 2024-10-25
refactor(sqlite): rename Sqlite::new() to Sqlite::in_memory()

open-schnick:
one has to read the new function to understand that the default behavior is in-memory


carllerche (approved):
This is reasonable, thanks.

PR #26, opened 2024-10-25
refactor(sqlite): fix some warnings

open-schnick:


src/db/sqlite/src/lib.rs
```diff
@@ -70,17 +70,13 @@ impl Driver for Sqlite {
         let mut stmt = connection.prepare(&sql_str).unwrap();
 
         if ty.is_none() {
-            let exec = match sql {
-                sql::Statement::Update(u) if u.pre_condition.is_some() => false,
-                _ => true,
-            };
+            let exec = !matches!(sql, sql::Statement::Update(u) if u.pre_condition.is_some());
```
open-schnick, line 73: maybe one could rename the variable to make it express the actual meaning behind not executing the query in some cases

carllerche: sure!

open-schnick: actually i am not really sure what the code means ^^
we dont exec when the sql is not an `Statement::Update` or an `Statement::Update` without a `pre_condition`
renaming exec to something more meaningful would make the code more readable i think

carllerche: When `exec` is true, call `stmt.execute`. When `exec` is false, call `stmt.query`). (i.e true is call the execute method).

PR #34, opened 2024-10-26
Fixed path errors in examples for windows and added a concrete sqlite example

EmilLindfors:
This is some minor changes to make examples easier to run. 
They wouldnt compile on windows, so changed to PathBuf.

Also added a sqlite example, to explicitly show multi-db support.


carllerche (approved):
Thanks!

Rustin170506:
@EmilLindfors The failed tests were caused by the format. You need to reformat it.

carllerche:
I want to get the path stuff merged in. That is probably good stand-alone PR.

After thinking a bit more about the SQLite example, I think a better approach instead of duplicating it would be to make sure all examples use SQLite instead of DBs that have to be run separately. That will make it easier for them to run locally.

So, steps to get this merged:
1) Revert the new example in this PR.
2) Create a separate PR that switches all examples to use SQLite.

Thanks!

EmilLindfors:
@carllerche sounds good, example is reverted

PR #38, opened 2024-10-28
Refactored examples

EmilLindfors:
I've switched to examples into using sqlite as proposed by @carllerche. 

Futher, I'm proposing an refactored examples folder that would make it easier to run the examples from root, as well as run the examples using different databases using features. See the README file for the quickstart. I'm thinking that should make it quite easy for folks to get started.


EmilLindfors:
@carllerche I reworked the cli a bit to be able to generate new examples easier, i though that would fit with the overall codegen vibe of this repo. See the README in toasty-examples for more info.

carllerche:
Thanks for taking a stab at this. I am with you that it should be easier to build the examples. I don't think adding "knowledge" of examples in the CLI is the way forward. Maybe open an issue to discuss ideas? I suspect a `build.rs` file will be the best way forward.

PR #45, opened 2024-11-02
move dependencies to workspace Cargo.toml

joshka:
Makes it easier to ensure that versions are same throughout

Note also bumps crate versions to latest available versions.


carllerche (approved):
Thanks.

PR #46, opened 2024-11-02
Add color to cli help output

joshka:
This is just using the cargo styling for now - tweak for toasty perhaps?


bsodmike:
Like the tweaks, cheers!

PR #47, opened 2024-11-02
FIX: Changes to user name should match example

bsodmike:
The example uses the wrong user value, this corrects this.

I've also stubbed out printing the entire schema structure to STDOUT and added a note.


carllerche (approved):
Thanks!

PR #49, opened 2024-11-02
Add rustqlite example

bsodmike:
This also incorporates corrections in https://github.com/tokio-rs/toasty/pull/47

### DIFF
--- a/examples/hello-toasty/src/main.rs
+++ b/examples/hello-toasty/src/main.rs
@@ -1,5 +1,5 @@
 mod db;
-use std::path::PathBuf;
+use std::{env, path::PathBuf};
 
 use db::{Todo, User};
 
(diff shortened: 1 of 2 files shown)


carllerche:
Hey, thanks for the PR. The examples all use sqlite now. If it is about persisting the DB to a file, maybe that can be added as a CLI option to the existing example?

bsodmike:
> Hey, thanks for the PR. The examples all use sqlite now. If it is about persisting the DB to a file, maybe that can be added as a CLI option to the existing example?

Good idea, let's do that :)

bsodmike:
@carllerche I opted for grabbing the CLI arg directly instead of using Clap etc.  Cheers!

examples/hello-toasty/src/main.rs
```diff
@@ -10,16 +10,27 @@ fn assert_sync_send<T: Send>(_: T) {}
...
+    let driver = if args.len() > 1 && args[1] == flag {
+        // Persist to local file
+        let filename = "toasty.db3";
+        let file_path = format!("./{}", filename);
+        let file = PathBuf::from(file_path.as_str());
+        Sqlite::open(file).unwrap()
+    } else {
+        // Use the in-memory sqlite driver
+        Sqlite::in_memory()
+    };
```
bit-aloo, line 26:
```suggestion
    const PERSIST_FLAG: &str = "--persist-to-file";
    const DB_FILENAME: &str = "toasty.db3";
    let driver = if args.get(1).map_or(false, |arg| arg == PERSIST_FLAG) {
        let file_path = PathBuf::from(DB_FILENAME);
        Sqlite::open(file_path).expect("Failed to open SQLite file")
    } else {
        Sqlite::in_memory()
    };
```

carllerche:
Thanks again for the PR. I'm going to close this in favor of #77 which provides a simple way to pick any supported DB backend in the examples. I'm open to considering a CLI arg instead of an env variable, but that can be a follow up PR.

PR #55, opened 2024-12-31
Allow self-referential BelongsTo relations

TannerRogalsky:
The BelongsTo has no indirection so for self-referential relations fails to compile due to an infinitely sized struct. This checks if the model and the relation share the same target and, for BelongsTo relations, adds a Box to the property type.

### DIFF
--- a/src/codegen/src/model/body.rs
+++ b/src/codegen/src/model/body.rs
@@ -177,8 +177,14 @@ impl<'a> Generator<'a> {
                 FieldTy::BelongsTo(rel) => {
                     let name = self.field_name(field);
                     let ty = self.model_struct_path(rel.target, 0);
-                    quote! {
-                        #name: BelongsTo<#ty>,
+                    if self.model.id == rel.target {
+                        quote! {
+                            #name: Box<BelongsTo<#ty>>,
+                        }
+                    } else {
+                        quote! {
+                            #name: BelongsTo<#ty>,
+                        }
                     }
                 }
                 FieldTy::Primitive(..) => {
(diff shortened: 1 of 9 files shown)


src/core/src/schema/verify.rs
```diff
@@ -45,7 +45,9 @@ impl Verify<'_> {
 
                 if let Some(belongs_to) = field.ty.as_belongs_to() {
                     assert_ne!(belongs_to.target, ModelId::placeholder());
-                    assert_ne!(belongs_to.pair, FieldId::placeholder());
+                    if model.id != belongs_to.target {
+                        assert_ne!(belongs_to.pair, FieldId::placeholder());
+                    }
```
TannerRogalsky, line 50: I wasn't really sure what this assert is protecting against so gated it behind the same check. There's likely a better solution.

src/codegen/src/model/body.rs
```diff
@@ -177,8 +177,14 @@ impl<'a> Generator<'a> {
                 FieldTy::BelongsTo(rel) => {
                     let name = self.field_name(field);
                     let ty = self.model_struct_path(rel.target, 0);
-                    quote! {
-                        #name: BelongsTo<#ty>,
+                    if self.model.id == rel.target {
+                        quote! {
+                            #name: Box<BelongsTo<#ty>>,
```
carllerche, line 182: Interesting, it might be worth always allocating in `BelongsTo`, especially since the structs might be big and not always loaded.

src/core/src/schema/verify.rs
carllerche: Is this assert being hit? `pair` should never be equal to `FieldId::placeholder()` once the schema is built. If the assert is being hit there is probably a bug.

carllerche (changes requested):
Hi! Thanks for submitting the PR. Sorry for the review delay. I was on vacation.

I agree that supporting self-referential belongs to is needed. I have a few requests before this gets merged.

First, instead of an example, could you add a test in the `tests/client` directory?

For the implementation strategy, I left an inline note, but I am wondering if having `value` in `BelongsTo<T>` always be boxed would be a better idea. It would simplify the code gen (avoiding having to conditionally box the field) and the structure sizes would be smaller.

what do you think?

TannerRogalsky:
I think this library is a little too early days for me to use even as an experiment. Best of luck!

PR #66, opened 2025-02-13
Adds a PostgreSQL client.

claymcleod:
This PR adds an initial PostgreSQL implementation. The goal was to introduce this on-par with what has been done with sqlite. Along the way, I added various methods that were needed or I felt like would be needed in the future.


Cargo.toml
```diff
@@ -43,6 +46,8 @@ serde = { version = "1.0.214", features = ["derive"] }
 serde_json = "1.0.132"
 std-util = { path = "src/std-util" }
 syn = "2.0.86"
+tokio = { version = "1.18", features = ["full"] }
```
claymcleod, line 50: Just pulled out the dependency from each crate into the workspace here.

src/core/src/stmt/value.rs
```diff
@@ -266,3 +266,37 @@ impl From<ValueRecord> for Value {
         Value::Record(value)
     }
 }
+
+#[cfg(feature = "postgres")]
+mod postgres_impl {
+    use super::*;
+    use postgres::types::accepts;
```
claymcleod: I'll come back and unify these imports (and any others).

tests/client/Cargo.toml
```diff
@@ -6,14 +6,8 @@ publish = false
...
 default = ["sqlite"]
-sqlite = [
-    "dep:toasty-sqlite",
-]
-dynamodb = [
-    "dep:toasty-dynamodb",
-    "dep:aws-config",
-    "dep:aws-sdk-dynamodb",
-]
+sqlite = ["dep:toasty-sqlite"]
```
claymcleod: Ignore these for now, I'll remove them.

carllerche:
Awesome, let me know when this is ready for review 👍

claymcleod:
This is ready for an initial review. You can do so with the following command:

```
DATABASE_URL="postgres://postgres@localhost/toasty" cargo run --release --bin example-hello-toasty-pgsql
```

Cargo.toml
```diff
@@ -10,13 +10,15 @@ members = [
...
     "src/db/sqlite",
+    "src/db/pgsql",
 
     # General utilities.
     "src/std-util",
 
     # Examples
     "examples/composite-key",
     "examples/hello-toasty",
+    "examples/hello-toasty-pgsql",
```
carllerche: This is fine for the PR, but we should figure out how to avoid copying examples for each new DB type. Maybe feature flags or something in the example.

src/core/Cargo.toml
```diff
@@ -8,8 +8,13 @@ publish = false
 anyhow.workspace = true
 async-trait.workspace = true
 indexmap.workspace = true
+postgres = { workspace = true, optional = true }
```
carllerche: The core crate should not include any DB driver dependencies.

src/core/src/stmt/value.rs
```diff
@@ -261,3 +261,32 @@ impl From<ValueRecord> for Value {
         Value::Record(value)
     }
 }
+
+#[cfg(feature = "postgres")]
```
carllerche: This should not be added to the core crate. Feel free to ping me to discuss strategies for keeping this in the driver crate.

carllerche (changes requested):
Looks good. I left a note about adding a PostgreSQL feature flag to the `core` crate. We should figure out how to move that code to the PG driver crate and keep `core` free of anything driver specific.

Also, how can we include PostgreSQL tests in CI?

Cargo.toml
claymcleod: I went ahead and resolved it.

src/core/src/stmt/sql/drop_table.rs
```diff
@@ -0,0 +1,43 @@
...
+            name: Name::from(&table.name[..]),
+            if_exists: false,
+        }
+        .into()
+    }
+
+    /// Drops a table if it exists.
+    ///
+    /// This function _does_ add an `IF EXISTS` clause.
+    pub fn drop_table_if_exists(table: &Table) -> Statement {
```
claymcleod, line 30: On something like this, I'd _much_ rather go with a builder-type pattern for `DropTable` to make this easier to use for the statement—perhaps add some convenience methods on `PostgreSQL` instance, but that would be about it. However, I wanted to follow the same style you've been using here, as I think it would need a wholesale switch over to builders for statements to not be confusing. I decided not to do that for this PR, but I'd like to know your thoughts on that.

src/core/src/stmt/value.rs
claymcleod: This should be resolved via a newtype, let me know if you don't like that strategy and we can try something else.

claymcleod:
Assuming the tests pass, this should be ready for review again.

There is the bigger question about how tests and examples should be run for all platforms/databases. I'd prefer to take that on in a future PR if that's amenable to you. For now, I've added a test for PostgreSQL for the one example that does support it (`example-hello-toasty`).

examples/cratehub/src/db/package.rs
```diff
@@ -17,6 +17,15 @@ impl Package {
             super::user::User::filter(super::user::User::ID.eq(&self.user_id)).into_select(),
         )
     }
+    pub async fn get_by_user_id(
```
claymcleod, line 20: I think this was the byproduct of me running `scripts/gen-examples` locally when I was testing. It seems like perhaps it wasn't run on some other change that affected this while merged in. I'm happy to just knock it out here if you'd like, but this is not a change I introduced intentionally.

tests/client/src/db.rs
```diff
@@ -3,5 +3,3 @@ pub mod dynamodb;
 
 #[cfg(feature = "sqlite")]
 pub mod sqlite;
-
-use super::*;
```
claymcleod, line 7: I was getting some _errors_ (?) when running this locally. Perhaps it's related to the recent 1.85 rollout? I'm not sure. Regardless, I felt like making these explicit below couldn't hurt.

examples/cratehub/src/db/package.rs
carllerche: Huh... There is some non-determinism in the code generator. I might know where. I will investigate that later.

src/core/src/stmt/sql/drop_table.rs
carllerche: Thanks for following the style. We can investigate a builder-style API in another PR. I'm generally fine adding builders.

carllerche (approved):
LGTM, thanks.

PR #67, opened 2025-02-15
Introduction of `clippy` to CI, related cleanups.

claymcleod:
This PR introduces `cargo clippy` to the CI and fixes the associated lint issues. I've tagged things I think require further review from @carllerche.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,20 @@ jobs:
         - name: cargo fmt
           run: cargo fmt --all --check
 
+    clippy:
+        name: cargo clippy
+        runs-on: ubuntu-latest
+        steps:
+        - uses: actions/checkout@v4
+        - uses: dtolnay/rust-toolchain@stable
+          with:
+            components: clippy
+        - uses: Swatinem/rust-cache@v2
+          with:
+            save-if: ${{ github.ref == 'refs/heads/main' }}
+        - name: cargo clippy
+          run: cargo clippy --workspace --all-features
+
     test-all-databases:
         needs: check
         name: Run tests with all databases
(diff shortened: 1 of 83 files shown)


src/codegen/src/model/relation.rs
```diff
@@ -201,11 +201,7 @@ impl Generator<'_> {
             quote!(#target_struct_name)
         };
 
-        let relation_struct = if field.ty.is_belongs_to() {
-            quote!(#relation_struct_name<'a>)
-        } else {
-            quote!(#relation_struct_name<'a>)
-        };
+        let relation_struct = quote!(#relation_struct_name<'a>);
```
claymcleod, line 204: This appears to be something that might have made sense to branch on but we never go around to it?

.github/workflows/ci.yml
```diff
@@ -27,6 +27,20 @@ jobs:
         - name: cargo fmt
           run: cargo fmt --all --check
 
+    clippy:
```
claymcleod, line 30: Unclear on if it would be better to just replace `cargo check`. What do you think?

src/core/src/stmt/value_stream.rs
```diff
@@ -53,6 +53,7 @@ impl ValueStream {
         }
     }
 
+    #[allow(clippy::should_implement_trait)]
```
claymcleod, line 56: I looked at this a bit closer, but I'm not sure what to do with it. I have not yet dug into the API design, some I'm going to leave it as-is for now.

claymcleod:
This is ready for review @carllerche.

src/codegen/src/model/relation.rs
carllerche: Not sure, but I am pretty sure that code path will go away soon anyway.

.github/workflows/ci.yml
carllerche: This seems fine.

Issue #69, opened 2025-02-21
SQL non standard issues


alelopezperez:
From what i've investigated it seems that there a some particularies in the serialization of a sql statement, that come into conflict with MySql(I'm developing support for MySql)

From what I gathered
~~1) Quoted table names from `toasty/src/core/src/stm/sql/serialize.rs` this is not supported in MySql~~ FIXED
```rust
  fn ident_str(&mut self, ident: &str) -> fmt::Result {
        write!(self.dst, "\"{ident}\"")?;
        Ok(())
    }
}
```
2) Since toasty uses TEXT type for ID, in MySQL it should be sized
`CREATE TABLE users (id TEXT, name INTEGER, email TEXT, moto TEXT, PRIMARY KEY (id));` => ERROR 1170 (42000) at line 1: BLOB/TEXT column 'id' used in key specification without a key length

It should be something like this
`CREATE TABLE usuarios (id TEXT, name INTEGER, email TEXT, moto TEXT, PRIMARY KEY (id(11)));`

3) RETURNING clause in mysql only works for INSERT, DELETE and REPLACE but not for UPDATE
[medium_article_evidence](https://dev.to/darkmavis1980/the-returning-clause-in-sql-11k2)
[stackoverflow_question](https://stackoverflow.com/questions/74329452/return-row-after-update-in-mariadb)
[mariadb_docs](https://mariadb.com/kb/en/replacereturning/)

alelopezperez:
They are many more places were wrapping of table names occur, in `toasty/src/core/src/stm/sql/serialize.rs` from what i gathered are the **select** **insert** **delete** and **update** from `impl<T: Params> Formatter<'_, T> `

carllerche:
Thanks. I recently refactored, which should make it easier to define per-db column types based on the specific database driver.  Also, the SQL serializer intends to support configuration options to handle this kind of case. More work will probably be needed to support MySQL and other SQL divergences.

alelopezperez:
Thanks for the clarification! I'll keep an eye on it and keep my branch updated

carllerche:
Do you want to take a stab at fixing quoting in the serializer? Add some config options to the [serializer struct](https://github.com/tokio-rs/toasty/blob/main/src/core/src/stmt/sql/serialize.rs#L17) and update the serialization code to respect them?

alelopezperez:
I'm interested, Yes, let's start with quoiting of the tables names.

carllerche:
[...]

PR #74, opened 2025-03-04
Add config options for Serializer  and Formater 

alelopezperez:
Added two config options for quoting table and column name in both the Serializer and Formatter; and change the corresponding code to respect the configs.

Solves issue  **1** from #69

### DIFF
--- a/src/core/src/stmt/sql/serialize.rs
+++ b/src/core/src/stmt/sql/serialize.rs
@@ -5,7 +5,7 @@ use crate::schema::db;
 use crate::stmt::Statement as DataStatement;
 
 use std::{
-    fmt::{self, Write},
+    fmt::{self, Display, Write},
     ops::Deref,
 };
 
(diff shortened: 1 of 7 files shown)


src/core/src/stmt/sql/serialize.rs
```diff
@@ -167,7 +191,10 @@ impl<T: Params> Formatter<'_, T> {
         write!(self.dst, " ON ")?;
 
         let table = self.schema.table(stmt.on);
-        self.ident_str(&table.name)?;
+
+        if self.quoted_table_names {
+            self.ident_str(&table.name)?;
+        }
```
carllerche: I don't see anything else. Would no table name be included at all in the fallback case?

carllerche (commented):
Thanks for the PR. What do you think of creating methods for `table_name` and `column_name` that encapsulate the quote / no quote condition in one place? Would that work? I saw some spots where it seemed like the `else` case was left out. Was that intentional?

src/core/src/stmt/sql/serialize.rs
alelopezperez: You are completely right, i did not put the else case so that would be an error

alelopezperez:
So what I'm thinking is that I could make a function that get called instead of ` write!(self.dst, "\"{}\"", ...)?;` and that function would be a method of the formatter; also it should run on the `ident()` function. Do you think this approach would be good? 

Also thanks for responding, is my first time getting involved in a project of this scale!

alelopezperez:
I think i pushed my commit, without re-basing with your recent commits I will correct it; meanwhile could you give me some feed on this solutiion.
I created a wrapper struct that implements the Display trait, there are some `write!` like this 
`write(dst, "INSERT {} INTO", name)` and others that are plainly `write!(dst, "{}",name)` 
```rust
struct MaybeQuote<'a> {
  value: &'a str,
  quote: bool
}
impl Display for MaybeQuote<'_>{
  fn fmt(...){
    if self.quoted {
            write!(f, "\"{}\"", self.value)
    } else {
            write!(f, "{}", self.value)
    }
  }
}
```

With the MaybeQuote wrapper struct it changes to
`write!(dst, "INSERT {} INTO", MaybeQuote::new(name, self.quote)"` and
`write!(dst, "{}",MaybeQuoter::new(name, self.quote))`
 we avoid cloning the name String and trying to make a new function that tries to handle different formatting cases

carllerche:
This seems fine to me, thanks.

Issue #78, opened 2025-03-13
libsql sqlite support


hbjydev:
It would be really cool to see support for `libsql` included in Toasty. I love the DX of using it currently, but I mainly use `libsql` for its performance and replication layer when running the server component.

I'd be happy to look at what would be necessary to add this. Maybe as a feature of the `sqlite` DB crate? (i.e. have the default implementation use `rusqlite` but allow use of `libsql` if you enable it)

hbjydev:
PR #79 raised with my draft of an implementation of this. :)

PR #79, opened 2025-03-13
feat: initial implementation of a libsql driver

hbjydev:
This PR aims to implement #78 by introducing a new `toasty-libsql` driver, supporting [`libsql`](https://github.com/tursodatabase/libsql).

It uses a lot of carbon-copy code from the `rusqlite` implementation, as minimal changes are needed to port between SQLite APIs and LibSQL by design.

Outstanding things to check/implement;

- [ ] I don't think `libsql::Database` is `Sync`, but `libsql::Connection` is. However, this means we have _one_ connection for the entire driver instance. I don't want to go as far as opening one connection per execution, but I'd like to look into some method of pooling the connections at some point.
- [x] ~~Token authentication to a LibSQL service needs a second parameter, but I don't know how to pass it just yet. I think some string manipulating/including the token in a query param then taking that parameter and passing it as the token might work.~~ **Implementing this via `impl From<libsql::Connection> for LibSQL` in the driver crate.**
- [ ] I'm doing a cast from `usize` to `i32` in the code for the driver, but I have no idea if that's sensible in the slightest; if there's a more 'safe' way to implement that type conversion I'd be very appreciative, but for now it got the code running.

The new driver currently works properly with all the example codebases, running them with `--features libsql` and setting `TOASTY_CONNECTION_URL` to `libsql::memory:` succeeds without any errors.

Let me know if I'm barking up the right tree here or not! :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,13 @@ jobs:
         runs-on: ubuntu-latest
         env:
             AWS_ENDPOINT_URL_DYNAMODB: http://localhost:8000
+        strategy:
+          fail-fast: false
+          matrix:
+            features:
+              - sqlite
+              - libsql
+              - dynamodb
         steps:
         - uses: actions/checkout@v4
         - uses: dtolnay/rust-toolchain@stable
@@ -54,9 +61,10 @@ jobs:
           with:
             save-if: ${{ github.ref == 'refs/heads/main' }}
         - name: Start dynamodb-local
+          if: matrix.features == 'dynamodb'
           run: sudo docker run --name dynamodb -d -p 8000:8000 amazon/dynamodb-local:latest -jar DynamoDBLocal.jar -port 8000
         - name: cargo test
-          run: cargo test --workspace --all-features
+          run: cargo test --workspace --no-default-features --features "${{ matrix.features }}"
 
     test-postgresql:
         needs: check
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ members = [
     "src/db/ddb",
     "src/db/sqlite",
     "src/db/pgsql",
+    "src/db/libsql",
 
     # General utilities.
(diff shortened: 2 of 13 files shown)


hbjydev:
I didn't see the multi-thread error locally... will investigate

carllerche:
Thanks for the PR. What are the pros / cons of libsql vs sqlite?

hbjydev:
> Thanks for the PR. What are the pros / cons of libsql vs sqlite?

So libsql primarily shines in a couple bigger places;
- Server component -- it has an HTTP implementation layer on top of the Sqlite codebase that lets you run it as a server instead of being stuck using it locally
- Full compatibility with SQLite because it's just a fork of it
- It has the ability to continuously snapshot itself to S3 if configured to do so
- Design-wise, it's focused on query latency and HA.

If you want Turso's argument, their [README](https://github.com/tursodatabase/libsql) lists a few more features it adds over plain SQLite, and their ['manifesto'](https://turso.tech/libsql-manifesto) provides the motivation behind building it.

hbjydev:
Also, r.e. this:
> Token authentication to a LibSQL service needs a second parameter, but I don't know how to pass it just yet.

I think I might implement `From<libsql::Database>` to let people bring their own DB configurations...

hbjydev:
Okay @carllerche I think I've figured out what breaks this... Apparently running `libsql` and `rusqlite` in the same thread has a bit of an issue attached; https://github.com/tursodatabase/libsql/issues/526

I'm not entirely sure how to resolve this in the context of `cargo test`, given it just runs them side-by-side.

Would you be okay with me splitting the test command to do `--feature libsql`, `--feature sqlite`, etc. individually? The tests do pass, this is just an issue with how `rusqlite` links its binary from what I can gather.

carllerche:
At this point, given the early development. I think I will focus on supporting sqlite directly In the future, it should be possible to support libsql as a third party crate.

PR #81, opened 2025-03-15
fix: postgresql connect

Caisin:


carllerche:
Thanks. I'm working towards getting the tests running under PG.

carllerche (approved):

carllerche:
I ended up bundling this change with #86 since CI isn't green here. Thanks again!

PR #83, opened 2025-03-17
Move table prefix logic to mapping

carllerche:


carllerche:
Whoops, I did this last week.

PR #95, opened 2025-03-28
Refactor sql serializer

carllerche:


crates/toasty-sql/src/serializer/flavor.rs
```diff
@@ -0,0 +1,33 @@
+use super::Serializer;
+
+use toasty_core::schema::db;
+
+#[derive(Debug)]
+pub(super) enum Flavor {
```
barsdeveloper, line 6: Won't this prevent external drive implementation? I am interested in DuckDB but it's not as popular as the DB trio everyone supports. I think it would be important to make it a design goal and allow for open extensions with trait base polymorphism that would allow an external crate to implement the ORM functionality of third party DBs. For instance this is my main criticism towards SeaORM, it uses everywhere enum based polymorphism which means that any driver must stay in the main repository and must be supported otherwise the orm cannot use that DB.

carllerche: No, it won't. `toasty-sql` is a utility library and not fundamental to implementing a driver. The individual SQL drivers can depend on `toasty-sql` to share code. A driver for a SQL implementation that isn't covered by `toasty-sql` can choose not to use it (or fork it to make needed changes or submit patches).

PR #96, opened 2025-04-10
WIP for MySQL Driver

alelopezperez:
This is an implemenation for the MySql driver heavily guided by the PostgreSQL drive. 

Currently I'm stopped by the the 2nd point on issue #69. Which is that Primary keys and Unique field need to be sized in MySQL, including in foreign keys.

in `crates/toasty-sql/src/serializer/stmt.rs` this two functions I could be able to solve it, I just curently don't understand how could i change it using the Flavor enum so it can format as ` PRIMARY KEY (id(255))` instead of `PRIMARY KEY (id)`
```rust
impl ToSql for ColumnsWithConstraints<'_> {
    fn to_sql<P: Params>(self, f: &mut super::Formatter<'_, P>) {
        let columns = Comma(&self.0.columns);

        if let Some(pk) = &self.0.primary_key {
            fmt!(f, columns ", PRIMARY KEY " pk);
        } else {
            fmt!(f, columns);
        }
    }
}

impl ToSql for &stmt::CreateIndex {
    fn to_sql<P: Params>(self, f: &mut super::Formatter<'_, P>) {
        let table_name = f.serializer.table_name(self.on);
        let columns = Comma(&self.columns);
        let unique = if self.unique { "UNIQUE " } else { "" };

        fmt!(
            f, "CREATE " unique "INDEX " self.name " ON " table_name " (" columns ")"
        );
    }
}
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
         runs-on: ubuntu-latest
         services:
           mysql:
-            image: mysql:8.0
+            image: mysql:9
             env:
               MYSQL_ROOT_PASSWORD: toasty
               MYSQL_USER: toasty
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,14 +67,15 @@ jobs:
           mysql:
             image: mysql:8.0
             env:
+              MYSQL_ROOT_PASSWORD: toasty
               MYSQL_USER: toasty
               MYSQL_PASSWORD: toasty
               MYSQL_DATABASE: toasty
             ports:
               - 3306:3306
-            options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+            options: --health-cmd "mysqladmin ping -h localhost" --health-interval 10s --health-timeout 5s --health-retries 5
         env:
-            TOASTY_CONNECTION_URL: mysql://toasty:toasty@localhost/toasty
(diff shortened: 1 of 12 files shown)


carllerche:
Hey, thanks for the PR. 

Regarding the sizing of strings in the PK, I have some thoughts on how to deal with it, but I think it would be best to punt until later. Lets get MySQL added and running in CI then we can work on improving the type support.

The short version is that I need to update Toasty such that the mapping of toasty types -> db types is database specific. I also need to start exposing attributes in the macro to allow the user to customize details (e.g. sizing of a string). I also need to add MySQL specifics to the `Capability` type to expose the limitation to the Toasty engine so it can catch issues.

Anyway, in short, could you add Mysql to CI?  You can probably copy how Postgresql runs in CI.

alelopezperez:
Thanks for the quick response!

So, I added mysql to the `tests/` and to CI, currently i commented out 
```yaml
test-mysql:
...
#        - name: cargo test
#          run: cargo test --workspace --no-default-features --features mysql
#        - name: cargo run --bin example-hello-toasty
#          run: cargo run --bin example-hello-toasty --features mysql
``` 
Since the test examples do have a the `TEXT` ID issue, I could make some other test just for the current version of the mysql driver implemantion and change does runs to use only those examples and tests.

carllerche:
Ok. I may merge this and fix it myself in a follow-up PR.

PR #101, opened 2025-04-28
fix: Unable to update data for multiple columns

### LINKED ISSUE #100: SQL Generation Issue for Multiple Column Updates
# The Problem
 
I'm trying to update multiple columns at once in Toasty, the SQL generator is failing to put commas between the SET assignments. This results in invalid SQL syntax:

```
UPDATE "users" SET "current_room_id" = ?1 "is_ready" = ?2 WHERE "id" = ?3 RETURNING "current_room_id", "is_ready";
```

This is the original code:

```rust
    let query = DbUser::filter(DbUser::FIELDS.id.eq(user_id))
        .update()
        .current_room_id(room_id.map(|s| s.to_string()))
        .is_ready(if is_ready { 1 } else { 0 });
```

The issue is that there should be a comma after "current_room_id" = ?1 but it's missing, causing the SQL syntax error.

# Root Cause

The bug is in Toasty's SQL serialization code. In `crates/toasty-sql/src/serializer/stmt.rs`, the implementation for formatting update statement assignments is using a space delimiter instead of the required comma:

https://github.com/tokio-rs/toasty/blob/7632b8ec3958f7412fdb4bd5fc8823bcf11096aa/crates/toasty-sql/src/serializer/stmt.rs#L255-L264

The correct implementation in the Toasty library should use a comma+space delimiter:

```rust
fmt!(f, Delimited(frags, ", "));  // Should use comma+space as delimiter
```

Losses:
Fix #100

### DIFF
--- a/crates/toasty-core/src/stmt/path_field_set.rs
+++ b/crates/toasty-core/src/stmt/path_field_set.rs
@@ -1,9 +1,9 @@
-use std::collections::HashSet;
+use indexmap::IndexSet;
 
 #[derive(Debug, Clone, Default, PartialEq, Eq)]
 pub struct PathFieldSet {
     // TODO: rewrite as a bitfield set
-    container: HashSet<usize>,
+    container: IndexSet<usize>,
 }
 
 impl PathFieldSet {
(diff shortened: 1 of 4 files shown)


carllerche:
Awesome! Any way to include a test? This seems like a big gap in the tests.

PR #112, opened 2025-05-29
update readme to align with the working code

alex-shapiro:
Small README update to make the code compile


carllerche (approved):
Good catch. Sorry for the delay. I am back.

PR #135, opened 2025-08-18
task: add an AlwaysTrue variant to stmt::Expr

mattyn:
This variant is useful when a Select statement has no filter expression, and evaluates the same as a literal value true.

### DIFF
--- a/crates/toasty-core/src/stmt/expr.rs
+++ b/crates/toasty-core/src/stmt/expr.rs
@@ -83,6 +83,9 @@ pub enum Expr {
 
     // TODO: get rid of this?
     DecodeEnum(Box<Expr>, Type, usize),
+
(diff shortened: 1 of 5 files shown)


carllerche:
I think I'm on board with having `filter: Option<Expr>` probably wasn't the right way. I don't think `AlwaysTrue` is correct either. After thinking about this. I think the way I would go is to define:

```rust
struct Filter {
    expr: Option<Expr>,
}
```

or something like that and use it directly on `Select` and other relevant statement structs

```rust
struct Select {
   // ... other fields

   pub filter: Filter,
}
```

Then, that struct can have any needed helper methods for dealing with the boilerplate of `None` vs. `true` etc...

That said, I'm going to close this PR. We can always open a new PR if we do want to pursue the above change.

PR #143, opened 2025-08-30
Add `serde::Serialize` support

joeydewaal:
This is the first step towards `serde` support. This implements `serde::Serialize` for `Id`, `BelongsTo`, `HasOne` and `HasMany` behind a new `serde` feature flag.

Now we can experiment with toasty in some web frameworks (axum,...).

A `not_loaded` helper function is also added so we can add `#[serde(skip_serializing_if = "HasMany::not_loaded")]` to structs.

### DIFF
--- a/crates/toasty/src/relation/belongs_to.rs
+++ b/crates/toasty/src/relation/belongs_to.rs
@@ -67,10 +67,10 @@ impl<T: fmt::Debug> fmt::Debug for BelongsTo<T> {
 }
 
 #[cfg(feature = "serde")]
-impl<T: serde::Serialize> serde::Serialize for BelongsTo<T> {
+impl<T: serde_core::Serialize> serde_core::Serialize for BelongsTo<T> {
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
-        S: serde::Serializer,
+        S: serde_core::Serializer,
     {
         self.value.serialize(serializer)
     }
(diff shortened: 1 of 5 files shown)


crates/toasty/src/relation/belongs_to.rs
```diff
@@ -23,6 +23,10 @@ impl<T: Model> BelongsTo<T> {
     pub fn get(&self) -> &T {
         self.value.as_ref().expect("association not loaded")
     }
+
+    pub fn not_loaded(&self) -> bool {
```
carllerche: not_loaded is a bit awkward. is_loaded would be better imo. I’m not super familiar with serde. Would that work?

carllerche (commented):
Generally looks good. I had a follow up question about not_loaded

crates/toasty/src/relation/belongs_to.rs
joeydewaal: Unfortunately not no. There is no way to negate the call (e.g `skip_serializing_if = "!HasMany::is_loaded"` doesn't work).

carllerche: Lets go with `is_unloaded()` then?

joeydewaal: Sure

joeydewaal:
I switched from `serde` to [`serde_core`](https://docs.rs/serde_core/1.0.221/serde_core/) which was released yesterday, this should help with compile times.

carllerche (approved):
Sorry for the delay, I am back!

PR #146, opened 2025-09-01
fix: Optimize O(N×M) association algorithm to O(N+M) for all relationship types

### LINKED ISSUE #141: Optimize N^2 association algorithm in HasMany relationship loading
## Problem

The current association algorithm in `crates/toasty/src/engine/exec/associate.rs` (lines 33-52) has O(N×M) time complexity when loading HasMany relationships. For each source record (N records), it iterates through all target records (M records) to find matches based on foreign key relationships.

## Current Implementation

```rust
// TODO: this is N^2, not super efficient.
for source_item in &mut source {
    let source_item = source_item.expect_record_mut();
    let mut associated = vec![];

    for target_item in &target {
        let target_item = target_item.expect_record();

        let [fk_field] = &pair.foreign_key.fields[..] else {
            todo!("composite keys")
        };

        if target_item[fk_field.source.index] == source_item[fk_field.target.index] {
            associated.push(target_item.clone().into());
        }
    }

    source_item[action.field.index] = stmt::Value::List(associated);
}
```

This nested loop approach becomes inefficient with larger datasets, especially when both source and target collections are large.

## Impact

- Performance degrades quadratically as the number of records increases
- Particularly affects queries with `.include()` that load related HasMany collections
- Could become a bottleneck in applications with large datasets

### Option 1: HashMap-based grouping (Recommended)
Pre-process the target records into a HashMap keyed by foreign key values:

```rust
// Build index: foreign_key_value -> Vec<target_records>
let mut target_by_fk: std::collections::HashMap<_, Vec<_>> = std::collections::HashMap::new();
for target_item in &target {
    let target_item = target_item.expect_record();
    let fk_value = &target_item[fk_field.source.index];
    target_by_fk.entry(fk_value.clone()).or_default().push(target_item.clone());
}

// Associate: O(N) lookup instead of O(N×M)
for source_item in &mut source {
    let source_item = source_item.expect_record_mut();
    let pk_value = &source_item[fk_field.target.index];
    let associated = target_by_fk.get(pk_value).cloned().unwrap_or_default();
    source_item[action.field.index] = stmt::Value::List(associated.into_iter().map(Into::into).collect());
}
```

**Complexity**: O(N + M) instead of O(N × M)

### Option 2: Database-level optimization
Consider pushing the association logic to the database layer with JOINs, though this may require more significant architectural changes.

### Option 3: Sorted arrays with binary search
If maintaining current data structures is preferred, sort target records by foreign key and use binary search, reducing complexity to O(M log M + N log M).

## Testing

The issue can be reproduced and performance improvements verified using the existing test `multiple_includes_same_model` with larger datasets.

## Additional Context

This optimization applies to the HasMany case. The BelongsTo case (lines 11-28) has a similar pattern but with early termination (`break`), making it less critical but could also benefit from similar optimization.

CreditWorthy:
This PR implements the HashMap-based optimization recommended in #141 to improve association loading performance.

## Changes

- Replace nested loop with HashMap-based grouping for HasMany associations (primary fix for #141)
- Apply the same optimization to BelongsTo and HasOne 
- Add Hash and Eq trait implementations to `stmt::Value` and related types to enable HashMap usage
- Modified `multiple_includes_same_model` test to support performance benchmarking with larger datasets
- Added Hash/Eq traits to supporting types for hashmap to work properly:
- `ValueEnum` (added Eq, Hash derives)
- `SparseRecord` (added Hash derive)
- `PathFieldSet` (manual Hash implementation for IndexSet compatibility)

## Performance Results

Tested with `multiple_includes_same_model` using larger dataset (500 users with 200 associations each):
- **Before**: 2.09 seconds for 100,000 associations
- **After**: 335ms for 100,000 associations

## Implementation Details

All three relationship types now use HashMap-based lookups:
- `BelongsTo`: HashMap<Value, ValueRecord> for one-to-one mapping
- `HasMany`: HashMap<Value, Vec<ValueRecord>> for one-to-many mapping  
- `HasOne`: HashMap<Value, ValueRecord> for one-to-one mapping

Fixes #141

new to contributing hope I did this right haha


crates/toasty-core/src/stmt/path_field_set.rs
```diff
@@ -48,3 +49,12 @@ impl FromIterator<usize> for PathFieldSet {
...
 }
+
+impl Hash for PathFieldSet {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.container.len().hash(state);
+        let mut items: Vec<_> = self.container.iter().copied().collect();
+        items.sort_unstable();
+        items.hash(state);
+    }
+}
```
carllerche: We can't really have this hash implementation. Instead, I would suggest improving PathFieldSet itself. Replace `IndexSet` with [`BitSet`](https://docs.rs/bit-set/latest/bit_set/struct.BitSet.html). `BitSet` implements `Hash`, so you should be good to go.

tests/tests/preload.rs
```diff
@@ -112,10 +112,107 @@ async fn multiple_includes_same_model(test: &mut DbTest) {
 
     let db = test.setup_db(models!(User, Post, Comment)).await;
 
-    // Create a user
+    // NOTE:
+    // 1. Temporarily change these values to (500, 100, 100)
+    // 2. Run with --nocapture to see timing
+    // 3. Revert to old algorithm and compare
+    let num_users = 10;
```
carllerche: Looks like you are trying to write a benchmark here. It probably makes more sense to set this up to use `cargo bench`: https://doc.rust-lang.org/cargo/commands/cargo-bench.html. Please remove the `println` at least to merge the change.

crates/toasty-core/src/stmt/value_record.rs
```diff
@@ -50,3 +50,11 @@ impl PartialEq for ValueRecord {
...
     }
 }
+
+impl Eq for ValueRecord {}
+
+impl std::hash::Hash for ValueRecord {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.fields.hash(state);
+    }
+}
```
carllerche, line 59: Do you know why `#[derive(Eq, Hash)]` didn't work here?

crates/toasty-core/src/stmt/value.rs
```diff
@@ -234,3 +235,29 @@ where
...
+            Value::U64(v) => v.hash(state),
+            Value::Id(v) => v.hash(state),
+            Value::String(v) => v.hash(state),
+            Value::Null => {}
+            Value::Record(v) => v.hash(state),
+            Value::List(v) => v.hash(state),
+            Value::SparseRecord(v) => v.hash(state),
+        }
+    }
+}
```
carllerche: Do you know why #[derive(Eq, Hash)] didn't work here?

crates/toasty-core/src/stmt/path_field_set.rs
zanville: tried it out, hope it looks good. Converting it to that did help with cleaning up the hash impl's all over the place, and let me use derive hash instead

tests/tests/preload.rs
zanville: yeah was alittle new to it haha, I added in an bench and an script to test the comparison and removed the old inline code, hope that's fine

crates/toasty-core/src/stmt/value_record.rs
zanville: I think at the time, when trying to implement hash. I noticed there was an trait conflict, with partial eq? same with value aswell, tho I noticed even with the bitset change I still have to impl hash to not have an conflict, only with this type tho

crates/toasty-core/src/stmt/value.rs
zanville: was an problem with partial eq, tho since the bitset change I can now derive eq and hash

crates/toasty-core/src/stmt/path_field_set.rs
```diff
@@ -21,13 +21,21 @@ impl PathFieldSet {
...
+        self.container.contains(val.into())
     }
 
+    // was looking into this and the only way to keep this impl is to allocate an vec
+    // tho I think it defeats the purpose of bitsets memory effficiency
     pub fn iter(&self) -> impl ExactSizeIterator<Item = usize> + '_ {
-        self.container.iter().map(Clone::clone)
+        let items: Vec<usize> = self.container.iter().collect();
+        items.into_iter()
     }
```
carllerche: Why? I looked at BitSet and it is true its Iter is not an "ExactSizeIterator" (which is unfortunate, since it can implement it). BitSet *does* provide `len()`, so could you define your own `Iter` struct here that holds the bit_set::Iter as well as impl `len()`?

zanville: that's a good point I didn't think about doing it like that

CreditWorthy:
> Thanks! Sorry for the delay. I am back now and should be more responsive.
> 
> I left one comment about the BitSet ExactLengthIterator thing. Also, (not a big deal) but in the future, if you want to add comments / questions about your PR inline, you can leave PR comments of your own. That way, they stay in the PR and don't get merged in.

oh ok awesome never tried doing that before I will note that for the future thank you

carllerche (approved):
Thanks!

PR #174, opened 2025-11-12
Add support for specifying a different database name for fields

### LINKED ISSUE #153: Feature Request: Add Field Attribute for Mapping to Different Database Column Names
Currently, the model mapping functionality relies on the model's field names directly corresponding to the database column names. The existing field attributes (#[key], #[auto], #[unique], #[index]) are very useful for defining key constraints and indexing, but they do not address the scenario where the desired model field name differs from the legacy or externally mandated database column name.

This limitation forces developers to name their model fields exactly as they appear in the database, which can lead to violations of application-level naming conventions (e.g., using snake_case in the database vs. camelCase in the model) and reduce the clarity and expressiveness of the model code.

I propose the introduction of a new field attribute that allows developers to explicitly specify the corresponding database column name for a model field. A good candidate for this attribute would be something like #[column("...")] or #[map("...").

This would allow the mapping logic to look for this attribute first and, if present, use the specified name to map to the database column. If the attribute is not present, it would fall back to the current behavior of using the field's name directly.

Example:

Let's say we have a database table users with a column named user_id. In our application code, we would prefer to use id for better readability. With the proposed attribute, we could define our model like this:

``` rust
struct User {
    #[key]
    #[column("user_id")] // New attribute
    id: i32,

    #[column("user_name")]
    name: String,
    
    // Other fields...
}
```

pikaju:
Adds custom column names with one of the following syntax options:

```rust
#[column("name")]
#[column(type = "type")]
#[column("name", type = "type")]
#[column(type = "type", "name")]
```

Removes the `#[db(type)]` attribute.

You can also either specify a varchar type like before:
```rust
#[column(type = varchar(5))]
```
or a custom type with a raw string:
```rust
#[column(type = "jsonb")]
```

Also adds two tests to check if the column was really renamed.

Closes #153

### DIFF
--- a/tests/tests/field_column_name.rs
+++ b/tests/tests/field_column_name.rs
@@ -17,9 +17,12 @@ async fn specify_custom_column_name(test: &mut DbTest) {
     let u = User::create().name("foo").exec(&db).await.unwrap();
     assert_eq!(u.name, "foo");
 
+    let mut filter = std::collections::HashMap::new();
+    filter.insert("id".to_string(), toasty_core::stmt::Value::from(u.id));
+
     // Make sure the column has actually been renamed to my_name in the underlying database.
     assert_eq!(
-        test.get_raw_column_value::<String>("users", "my_name", Default::default())
+        test.get_raw_column_value::<String>("users", "my_name", filter)
             .await
             .unwrap(),
         "foo"
@@ -27,6 +30,10 @@ async fn specify_custom_column_name(test: &mut DbTest) {
 }
 
 async fn specify_custom_column_name_with_type(test: &mut DbTest) {
+    if test.capability().storage_types.varchar.is_none() {
+        return;
+    }
+
     #[derive(toasty::Model)]
     struct User {
         #[key]


crates/toasty-core/src/schema/db/ty.rs
```diff
@@ -15,6 +15,9 @@ pub enum Type {
     Text,
 
     VarChar(u64),
+
+    /// User-specified unrecognized type
+    Custom(&'static str),
```
carllerche: We *probably* want this to be String to allow runtime specification of custom types. I don't think allocations will be an issue.

pikaju: Okay, I changed it to `String`

crates/toasty-core/src/schema/builder/table.rs
```diff
@@ -364,22 +369,24 @@ impl BuildTableFromModels<'_> {
...
         };
 
         let column = db::Column {
             id: ColumnId {
                 table: self.table.id,
                 index: self.table.columns.len(),
             },
-            name,
+            name: name.app_name.clone(),
+            storage_name,
```
carllerche, line 389: I'm OK merging the PR with this, but as a follow up, I don't think we need to track `app_name` here. `app_name` is the app-level field name. `db::Column` is the storage name. If there is something I am missing, let me know.

Issue #147, opened 2025-09-15
Get an `Id<T>` without going through toasty


joeydewaal:
Currently there is no way create an `Id<T>` without going through toasty. By for example creating, updating or selecting the model.

E.g. if you have a `user` table with an `id INT PRIMARY KEY` field, it's not possible to do
```rust
User::get_by_id(&db, 1i32)
```

You have to get an `Id<User>` somehow.

joeydewaal:
Closing this since #334 has been merged

PR #187, opened 2025-11-16
Allow constructing Id<T> from raw integers

zxq82lm:
## PR objective
Address the issue https://github.com/tokio-rs/toasty/issues/147.

## Testing

Unit test:
```
cargo test numeric_id_expr
```

### DIFF
--- a/crates/toasty-core/src/stmt/id.rs
+++ b/crates/toasty-core/src/stmt/id.rs
@@ -68,22 +68,32 @@ impl Id {
 
     pub fn to_primitive(&self) -> stmt::Value {
         match &self.repr {
-            Repr::Int(_) => todo!(),
+            Repr::Int(id) => stmt::Value::String(id.to_string()),
             Repr::String(id) => id.clone().into(),
         }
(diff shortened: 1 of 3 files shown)


crates/toasty-core/src/stmt/id.rs
```diff
@@ -68,22 +68,32 @@ impl Id {
 
     pub fn to_primitive(&self) -> stmt::Value {
         match &self.repr {
-            Repr::Int(_) => todo!(),
+            Repr::Int(id) => stmt::Value::String(id.to_string()),
```
carllerche, line 71: Is there a reason you are converting an Int ID to a string?

zxq82lm: I’m doing the Int-backed Id → Value::String conversion because the current pipeline maps Type::Id to a string column (`stmt_ty_to_table` does Type::Id(_) => Type::String), and the drivers already serialize IDs as text. If you’d prefer a different representation I’m happy to adjust.

PR #192, opened 2025-11-18
Add auto increment keys and auto fields with UUID v4 or v7

pikaju:


crates/toasty-core/src/stmt/insert.rs
```diff
@@ -11,6 +11,13 @@ pub struct Insert {
 
     /// Optionally return data from the insertion
     pub returning: Option<Returning>,
+
+    /// Statements to execute after the INSERT completes
+    ///
+    /// These statements can reference fields from the inserted record using
+    /// `Expr::ParentField`. They execute after the parent INSERT but do not
+    /// affect the returned results.
+    pub then: Vec<Statement>,
```
carllerche: I know it was in the design doc. I would punt on it for the initial PR. I don't think any current test will require this to pass.

tests/tests/field_auto.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    for i in 1..10 {
+        let u = Foo::create()
+            .bar(Bar::create())
+            .bar(Bar::create())
+            .exec(&db)
+            .await
+            .unwrap();
+        assert_eq!(u.id, i);
+        assert_eq!(u.bars.get()[0].foo.get().id, i);
```
carllerche: foo.get() should *not* work here since the foo association is not eager loaded. You will need to issue a query to load it. I don't remember the API, maybe check another test file

PR #201, opened 2025-11-27
Add date/time times using `jiff`

pikaju:
Adds the following new Toasty value types, hidden behind a `jiff` feature flag:

```rust
    /// An instant in time represented as the number of nanoseconds since the Unix epoch.
    Timestamp(jiff::Timestamp),

    /// A time zone aware instant in time.
    Zoned(jiff::Zoned),

    /// A representation of a civil date in the Gregorian calendar.
    Date(jiff::civil::Date),

    /// A representation of civil “wall clock” time.
    Time(jiff::civil::Time),

    /// A representation of a civil datetime in the Gregorian calendar.
    DateTime(jiff::civil::DateTime),
```

See the docs markdown file for more information.

### DIFF
--- a/tests/tests/ui/relation_has_one_requires_attr.stderr
+++ b/tests/tests/ui/relation_has_one_requires_attr.stderr
@@ -11,6 +11,6 @@ error[E0277]: the trait bound `toasty::HasOne<Option<Profile>>: toasty::stmt::Pr
             i64
             i8
             isize
-            std::string::String
-            toasty::stmt::Id<T>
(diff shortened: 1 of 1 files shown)


joeydewaal:
Sorry to bother but should the default `Zoned` type for Postgres and MySQL be `TIMESTAMP`? The maintainer of jiff pointed out that it kinda is a footgun:

https://github.com/launchbadge/sqlx/issues/3487#issuecomment-2638197946

PR #204, opened 2025-11-28
Adds docs for expression statements, tests for simplifications

claymcleod:
This PR represents (a) code documentation and (b) new tests I added while exploring the simplification phase of the engine.

### DIFF
--- a/crates/toasty-core/src/stmt/expr_any.rs
+++ b/crates/toasty-core/src/stmt/expr_any.rs
@@ -2,8 +2,8 @@ use super::Expr;
 
 /// Returns `true` if any item in a collection evaluates to `true`.
 ///
-/// [`ExprAny`] evaluates its inner expression and returns `true` if at least one
-/// item is truthy. Returns `false` for an empty collection.
+/// Evaluates its inner expression and returns `true` if at least one item is
+/// truthy. Returns `false` for an empty collection.
 ///
 /// # Examples
 ///


crates/toasty/src/engine/simplify/lift_in_subquery.rs
```diff
@@ -49,13 +49,16 @@ impl Simplify<'_> {
         maybe_expr
     }
 
-    /// Optimizes queries by lifting BelongsTo relation constraints out of subqueries when possible.
+    /// Optimizes queries by lifting BelongsTo relation constraints out of
```
claymcleod: This is due to my editor autowrapping, happy to revert if it's not desired.

crates/toasty/src/engine/simplify/association.rs
```diff
@@ -70,3 +72,291 @@ impl Simplify<'_> {
...
+                BelongsTo, Field, FieldId, FieldName, FieldPrimitive, FieldTy, ForeignKey,
+                ForeignKeyField, HasMany, Index, IndexField, IndexId, Model, ModelId, PrimaryKey,
+            },
+            db::{IndexOp, IndexScope},
+            Builder, Name,
+        },
+        stmt::{Association, Expr, ExprInSubquery, Path, Query, SourceModel, Type, Value},
+    };
+
+    struct UserPostSchema {
```
carllerche, line 92: I want to avoid hand-writing schemas if at all possible. In tests, it should be possible to use the macros to define a schema, then use that?

If you want to extract the tests to a different PR, the docs can be merged as is.

Issue #185, opened 2025-11-15
Validate storage type requested by user with `#[column(type = ...)]`


carllerche:
If the rust model type is a String and the storage type is an int, that probably doesn't make sense and should be an error at compile time?

carllerche:
## Issue Confirmed

I've investigated this issue and can confirm it's a real problem. The `#[column(type = ...)]` attribute currently accepts any storage type without validating compatibility with the Rust field type.

## Current Behavior

When a user specifies a storage type via `#[column(type = ...)]`, Toasty accepts it without validation, even if the storage type is incompatible with the Rust type. For example:

```rust
#[derive(toasty::Model)]
struct User {
    #[key]
    #[auto]
    id: Id<Self>,

    // This compiles but doesn't make sense - String with int storage
    #[column(type = "int")]
    name: String,
}
```

This code **compiles successfully** at the macro expansion stage, but the type mismatch could cause runtime errors or data corruption depending on how drivers handle the mismatch.

## Root Cause

The issue is in `crates/toasty-core/src/schema/db/ty.rs:86-115`, specifically in the `Type::from_app` method:

```rust
pub fn from_app(
    ty: &stmt::Type,
    hint: &Option<Type>,
    db: &driver::StorageTypes,
) -> Result<Type> {
    match hint.clone() {
        Some(ty) => Ok(ty),  // ← No validation! Just returns the hint
        None => match ty {
            // ... proper mapping logic when no hint is provided
        }
    }
}
```

When a `#[column(type = ...)]` attribute is present, the hint is set and returned directly without checking if it's compatible with the Rust application type.

## Suggested Solutions

### Option 1: Compile-Time Validation (Recommended)

Add validation in `crates/toasty-codegen/src/schema/field.rs` when parsing the `#[column]` attribute. This would catch errors at compile time with clear error messages.

**Implementation approach:**
- After parsing the column type, validate it against the field's Rust type
- Create a compatibility matrix (e.g., String → {Text, VarChar, Custom}, i64 → {Integer(8), Custom})
- Emit a compile error if the types are incompatible

**Pros:**
- Catches errors at compile time
- Clear, immediate feedback to developers
- No runtime overhead

**Cons:**
- Requires maintaining a compatibility matrix
- Need to handle edge cases like `Option<T>` wrappers

### Option 2: Runtime Validation

Enhance the `Type::from_app` method to validate the hint against the application type:

```rust
match hint.clone() {
    Some(storage_ty) => {
[...]

PR #208, opened 2025-11-29
feat: Add schema-time type validation with bool primitive support

brndnblck:
## Summary

Implements runtime type validation during schema building to catch type mismatches that can't be detected at macro expansion time, such as type alias shadowing. Also adds comprehensive bool primitive support across all database drivers.

## Problem Solved

Fixes issue #185 where type validation at macro time fails with type aliases.

Example that was problematic:
```rust
type String = u32;  // shadows built-in String

#[derive(toasty::Model)]
struct User {
    #[column(type = text)]
    name: String,  // This is actually u32, not String!
}
```

Macro-time validation cannot detect this because it sees the identifier `String` and assumes it's `std::String`. Schema-time validation catches this at runtime during schema building when the actual `stmt::Type` is resolved.

## Key Features

- **Schema-time validation** in `toasty-core/schema/verify.rs` that catches type mismatches at schema build time
- **Whitelist approach** for maintainability with explicit compatible type combinations
- **Type alias detection** that handles complex scenarios macros cannot resolve
- **Clear error messages** with type suggestions for better developer experience
- **Full bool primitive support** including proper SQLite INTEGER(0/1) value conversion
- **Comprehensive test coverage** for both validation logic and cross-driver integration

### Core Changes

- `crates/toasty-core/src/schema/verify.rs` - Added `verify_column_type_compatibility()` with comprehensive type checking
- `crates/toasty-core/src/stmt/value.rs` - Added bool value conversion support
- `crates/toasty/src/stmt/primitive.rs` - Added bool primitive with proper trait implementation ordering
- Database drivers updated for bool support across SQLite, PostgreSQL, and MySQL

### Test Coverage

- `tests/tests/field_column_type_validation.rs` - Tests for valid type combinations
- `tests/tests/schema_validation_test.rs` - Tests for validation error cases with clear error messages
- Integration tests updated across all database drivers

## Validation Logic

The implementation uses a whitelist approach that explicitly defines compatible type combinations:

- **String types**: `String`, `&str` compatible with `TEXT` columns
- **Integer types**: All signed/unsigned integer types with size validation
- **Bool types**: `bool` compatible with `BOOLEAN` columns (stored as INTEGER in SQLite)
- **UUID types**: `uuid::Uuid` compatible with `UUID` columns
- **Comprehensive validation**: Covers all primitive types with appropriate database storage types

## Benefits

- **Early error detection**: Catches type mismatches at schema build time rather than runtime
- **Type alias safety**: Handles complex scenarios that macro validation cannot
- **Clear diagnostics**: Provides helpful error messages with suggested compatible types
- **Zero runtime overhead**: Validation occurs only during schema construction
- **Extensible design**: Easy to add new type compatibility rules as needed


crates/toasty-codegen/src/schema/field.rs
```diff
@@ -157,6 +158,18 @@ impl Field {
             ));
         }
 
+        // Validate column type compatibility for primitive fields
+        if ty.is_none() && attrs.column.is_some() {
```
carllerche:
```suggestion
        if ty.is_none() {
```

`attires.column.is_some()` is checked in `validate_column_type_compatibility`.

brndnblck:
@carllerche I struggled with this one a bit but would welcome another pass and your feedback. 

Compile time validation wasn't really possible, but this does implement schema-time validation that fails on start at runtime vs. blowing up later on during execution or causing data loss. It's unfortunately less of a compile time catch and more of a fast-fail approach, but I think it still adds value.

`cargo build` - ✅ Compiles fine (no validation enforced)
`cargo run` - ❌ Panics at runtime in startup during schema building

Issue #215, opened 2025-11-30
[Tracking] Expression Simplification Rules


claymcleod:
This issue tracks the expression simplification rules to be implemented in Toasty's query engine.

## Boolean Algebra

| Simplification | Source | Status | Notes |
|----------------|--------|--------|-------|
| `not(not(x))` → `x` | [DataFusion](https://github.com/apache/datafusion/blob/main/datafusion/optimizer/src/simplify_expressions/expr_simplifier.rs) | #214 | Double negation elimination |
| `not(true)` → `false`, `not(false)` → `true` | [DataFusion](https://github.com/apache/datafusion/blob/main/datafusion/optimizer/src/simplify_expressions/expr_simplifier.rs) | #214 | Constant folding for `not` |
| `not(x = y)` → `x != y` | [DataFusion](https://github.com/apache/datafusion/blob/main/datafusion/optimizer/src/simplify_expressions/expr_simplifier.rs) | #214 | Negation of comparisons (all 6 ops) |
| `not(a and b)` → `not(a) or not(b)` | [DataFusion](https://github.com/apache/datafusion/blob/main/datafusion/optimizer/src/simplify_expressions/expr_simplifier.rs) | #216 | De Morgan's law |
| `not(a or b)` → `not(a) and not(b)` | [DataFusion](https://github.com/apache/datafusion/blob/main/datafusion/optimizer/src/simplify_expressions/expr_simplifier.rs) | #216 | De Morgan's law |
| `x and true` → `x` | [DataFusion](https://github.com/apache/datafusion/blob/main/datafusion/optimizer/src/simplify_expressions/expr_simplifier.rs) | (pre-existing) | AND identity law |
| `x and false` → `false` | [DataFusion](https://github.com/apache/datafusion/blob/main/datafusion/optimizer/src/simplify_expressions/expr_simplifier.rs) | #206 | AND dominance law (even if x is null) |
| `x or true` → `true` | [DataFusion](https://github.com/apache/datafusion/blob/main/datafusion/optimizer/src/simplify_expressions/expr_simplifier.rs) | #209 | OR dominance law (even if x is null) |
| `x or false` → `x` | [DataFusion](https://github.com/apache/datafusion/blob/main/datafusion/optimizer/src/simplify_expressions/expr_simplifier.rs) | #209 | OR identity law |
| `and(and(a, b), c)` → `and(a, b, c)` | — | (pre-existing) | AND flattening |
[...]

PR #229, opened 2025-12-02
feat: simplify is_null on non nullable fields

joeydewaal:
This simplifies is null expressions on a non nullable field and evaluates them to `false`.

This is also mentioned in #215 under NULL Handling.


crates/toasty/src/engine/simplify.rs
```diff
@@ -285,4 +288,12 @@ mod test {
             .build(app::Schema::default(), &Capability::SQLITE)
             .expect("empty schema should build")
     }
+
+    pub fn test_schema_with(models: &[Model]) -> toasty_core::Schema {
```
carllerche, line 292: As a follow up, this probably should be  `pub(crate)` with a `#[cfg(test)]`. I will track that separately though.

joeydewaal: Ah yes I'll open a pr. I assume `test_schema`, the function above `test_schema_with` should also have this?

joeydewaal: If so, I'll open a separate pr for it.

PR #233, opened 2025-12-03
feat: support smartpointers

joeydewaal:
This pr adds support for `Box`, `Rc` and `Arc` in model fields.

This is something people asked for in sqlx, so I thought I'd add it here as well.


joeydewaal:
I merged main locally which removed your (carl) commit, I hope that's not a problem.

carllerche:
@joeydewaal nope, thanks

PR #238, opened 2025-12-04
Add basic support for bigdecimal::BigDecimal

pikaju:
Unfortunately all DBs store `BigDecimal` as `TEXT` for now, as the Postgres encoding is more complicated than I thought. I want to add `rust-decimal` next which will work with Postgres' `NUMERIC`.


carllerche (approved):
Thanks :+1:

PR #251, opened 2025-12-07
feat: adds in list simplifications

claymcleod:
This PR implements all of the "in list" simplifications pointed out in #215.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,6 +69,7 @@ pretty_assertions = "1.4.1"
 proc-macro2 = "1.0.37"
 quote = "1.0.18"
 rand = "0.8.5"
+rust_decimal = "1.39.0"
 bigdecimal = "0.4.7"
 rusqlite = { version = "0.32", features = ["bundled"] }
 serde = { version = "1.0.214", features = ["derive"] }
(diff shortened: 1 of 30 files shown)


crates/toasty/src/engine/simplify/expr_or.rs
```diff
@@ -177,6 +182,86 @@ impl Simplify<'_> {
...
+    /// values into IN lists. Non-equality operands are preserved.
+    fn try_or_to_in_list(&self, expr: &mut stmt::ExprOr) -> Option<stmt::Expr> {
+        use std::collections::HashMap;
+
+        // Group equalities by their LHS expression
+        // Key: index into `lhs_exprs`, Value: list of RHS constant values
+        //
+        // TODO: this could be simplified to use `Expr` as the `HashMap` key
+        // directly if `Expr` ever implements `Hash`.
+        let mut lhs_exprs: Vec<stmt::Expr> = Vec::new();
```
claymcleod, line 201: I had to do this ugliness here because `Expr` doesn't implement `Hash`. Can it? I didn't want to take on that piece in this PR, but worth coming back to if it can.

carllerche: We will stick with this for now, but I don't think implementing Hash is the right way to do it. There needs to be a separate system of "equivalence" that goes deeper than equality / hash.

PR #252, opened 2025-12-08
feat: adds arithemetic operations to the IR

claymcleod:
This PR adds arithmetic operations to the IR along with some simplifications. I think it should really be a discussion ground as to whether supporting arithmetic representations is going to be useful in `toasty` down the line. I can see it being useful in things like precomputed columns and customized select statements where you want to do the operations within the database rather than in Rust. Open to thoughts.

### DIFF
--- a/crates/toasty-core/src/stmt/op_unary.rs
+++ b/crates/toasty-core/src/stmt/op_unary.rs
@@ -1,0 +1,23 @@
+use std::fmt;
+
+/// Unary operators for expressions.
+// TODO: consider pulling logical negation (`Not`) into this enum.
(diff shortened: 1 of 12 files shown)


crates/toasty-core/src/stmt/op_unary.rs
```diff
@@ -0,0 +1,23 @@
+use std::fmt;
+
+/// Unary operators for expressions.
+// TODO: consider pulling logical negation (`Not`) into this enum.
```
claymcleod, line 4: Notably, we should pull `Not` into this if we decide to go this route.

carllerche:
Closing due to inactivity

PR #255, opened 2025-12-11
Add support for `chrono`

joeydewaal:
This pr adds support for the `chrono` crate.

I prefixed the `jiff` values with `Jiff` for clarity.


carllerche:
Thanks. What does chrono offer over jiff? or is it more a question of better supporting apps that already use chrono?

joeydewaal:
It looks like chrono is being soft deprecated, so it might be less interesting for toasty to support it.

(there is no issue to reference, so from djc's blog)
image

PR #260, opened 2025-12-16
Add database connection pooling

pikaju:


crates/toasty-core/src/driver.rs
```diff
@@ -13,17 +13,29 @@ use std::{fmt::Debug, sync::Arc};
...
+        None
+    }
+}
 
-    /// Register the schema with the driver.
-    async fn register_schema(&mut self, schema: &Schema) -> crate::Result<()>;
+#[async_trait]
+pub trait Connection: Debug + Send + 'static {
+    /// Describes the driver's capability, which informs the query planner.
+    fn capability(&self) -> &'static Capability;
```
carllerche, line 32: You probably still want this on `Driver` no?

carllerche (approved):
LGTM, only question is why move `capability()` -> `Connection` instead of keeping it on `Driver`. It doesn't seem like a connection specific concern to me?

PR #299, opened 2026-01-29
feature: support enum field types

carllerche:
I am starting to work on designing this feature and will use this PR to track discussion and design.

[Rendered](https://github.com/tokio-rs/toasty/blob/enum-field-ty/docs/design/enums-and-embedded-structs.md)

## TODO

* [ ] How to update only part of a nested structure/enum.
* [ ] How to use embedded fields in queries

### DIFF
--- a/crates/toasty/src/engine/simplify/association.rs
+++ b/crates/toasty/src/engine/simplify/association.rs
@@ -77,7 +77,7 @@ impl Simplify<'_> {
 mod tests {
     use super::*;
     use crate as toasty;
-    use crate::Model as _;
+    use crate::model::Register;
     use toasty_core::{
         driver::Capability,
         schema::{app, app::FieldId, app::ModelId, Builder},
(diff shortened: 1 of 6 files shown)


carllerche:
> In this section, it shows an example where #[column("name")] on the enum yields a column named "name", not prefixed by the field name with the enum as a type. Was that intentional? It doesn't seem to match other examples using that attribute.

No, it should always be prefixed by the field name. Thanks for the catch.

> Regarding the discriminator, what do you think about supporting a string discriminator? 

Should be supported. I think you would just specify `#[column(type = "<string-type>")]` and do `#[column(variant = "foo")]`.

> Final thought for now: in DynamoDB it would often be preferable to represent enum fields as document-style values.

I'm thinking this will actually be a separate feature path. Something like:

```rust
#[derive(serde::Serialize, serde::Deserialize)]
enum Creature {
    Human { profession: String },
    Lizard { habitat: String },
}

#[serialize] // not sure what the actual attribute will be.
creature: Creature,
```

PR #308, opened 2026-02-10
Add {update/delete}_by_id snippets

joeydewaal:
Since updating and deleting models by id is very common I thought I'd add snippets for this.

Now it's possible to do:
```rust
User::update_by_id(user_id)
  .name(...)
  .exec(&db).await;

User::delete_by_id(&db, user_id).await;
```


carllerche:
Nice, just needs some tests.

Issue #378, opened 2026-02-26
Design: Connection-per-Db Instance


pikaju:
## Problem

1. **Read-your-writes**: Each `exec_plan()` acquires a fresh pool connection. Sequential operations on the same `Db` may hit different connections, so a write followed by a read may not see the write.

2. **Interactive transactions**: Transactions must pin to a single connection. The current model acquires per-plan, making multi-statement transactions impossible.

3. **Single execution task**: One spawned task per `Db`, but it acquires a new connection per plan.

## Current Architecture

```rust
Db {
    engine: Engine { schema, pool },
    in_tx: UnboundedSender<(Statement, oneshot::Sender)>,
    join_handle: JoinHandle<()>,
}
```

- `Db` is not `Clone`. All methods take `&self`.
- One channel + task per `Db`; the task calls `pool.get()` per plan.

### `Db` becomes `Clone`, each instance owns a connection + task

```rust
#[derive(Clone)]
pub struct Db {
    shared: Arc<Shared>,
    conn: Option<ConnHandle>,  // lazily acquired, None on clone
}

struct Shared {
    engine: Engine,
    pool: Pool,
}

struct Engine {
    schema: Schema,
    capability: &'static Capability,
}

struct ConnHandle {
    in_tx: UnboundedSender<ConnOp>,
    join_handle: JoinHandle<()>,
}
```

On first use, `Db` checks out a `PoolConnection` from the pool, spawns a dedicated tokio task that owns it, and stores the channel handle:

```rust
async fn connection(&mut self) -> Result<&ConnHandle> {
    if self.conn.is_none() {
        let mut connection = self.shared.pool.get().await?;
        let (in_tx, mut in_rx) = mpsc::unbounded_channel();

        let join_handle = tokio::spawn(async move {
            while let Some(op) = in_rx.recv().await {
                // execute op on connection, send result back via oneshot
            }
            // channel closed → connection drops → returns to pool
        });

        self.conn = Some(ConnHandle { in_tx, join_handle });
    }
    Ok(self.conn.as_ref().unwrap())
}
```

- **Clone**: Cloned instances get `conn: None`, will lazily acquire their own connection + task.
- **Drop**: `ConnHandle` drops → channel closes → task exits → `PoolConnection` returns to pool.

### All methods take `&mut self`

```rust
impl Db {
    pub async fn exec<M: Model>(&mut self, stmt: Statement<M>) -> Result<ValueStream> {
        let conn = self.connection().await?;
[...]

PR #379, opened 2026-02-26
Connection per Db

pikaju:
Implements #378


crates/toasty/src/db/pool.rs
```diff
@@ -86,9 +86,21 @@ impl Pool {
     }
 
     /// Returns the database driver's capabilities.
-    pub fn capability(&self) -> &Capability {
+    pub fn capability(&self) -> &'static Capability {
```
carllerche, line 89: Why return &'static Capability?

PR #380, opened 2026-02-26
DDB N+1 Select Fix

### LINKED ISSUE #338: DDB Support - Use attribute from GSI
When GSIs use projection type ALL, entities can be built from the result of querying the GSI. The current implementation will throw away the other fields, then issue a BatchGetItem for the PKs. 

Fix: Use the result of the GSI query to build the entities.

asiefker:
Previously, toasty would use the GSI to find keys, then issue a
BatchGetItem to build the item. With this change, the Item(s) will be
built directly from the GSI.

- **Add trace and logging to DDB driver.**
- **DDB - Use GSI directly if it has all columns**

Fixes #338

### DIFF
--- a/crates/toasty-core/src/driver.rs
+++ b/crates/toasty-core/src/driver.rs
@@ -5,7 +5,7 @@ mod response;
 pub use response::{Response, Rows};
 
 pub mod operation;
-pub use operation::Operation;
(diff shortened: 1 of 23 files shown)


asiefker:
Moved to assert! and simplified the DDB query building.

PR #391, opened 2026-03-02
DDB Client Creation Refactor

### LINKED ISSUE #388: Refactor DDB Client integration
1. Remove custom URL endpoint parsing - Use standard AWS SDK environment variables instead
  2. Share AWS SDK Client across connections - Move Client to Driver struct, clone when creating connections (performance optimization)
  3. Accept custom SdkConfig - Add constructors that take pre-configured SdkConfig for full AWS SDK flexibility

### Standard AWS Environment Variables:
  - AWS_ENDPOINT_URL_DYNAMODB is the standard way to configure DynamoDB endpoint (including DynamoDB Local)
  - dynamodb:// URL continues to work with toasty::db::Connect::new() for driver selection, but AWS SDK configuration comes from environment
  - Alternative: Create DynamoDB driver directly with custom config and pass to Db::builder().build(driver), bypassing URL-based configuration entirely

### Client Reuse (Performance):
  [Clients should be reused](https://docs.aws.amazon.com/sdk-for-rust/latest/dg/best-practices.html) because of connection pooling. We now create the client once in the Driver and clone it (cheap Arc clone) for each Connection.

### Custom Configuration Support:
SdkConfig supports [CredentialProviders](https://docs.aws.amazon.com/sdk-for-rust/latest/dg/credproviders.html) and other options (region, retries, timeouts, etc.). New API:
  - DynamoDb::from_env() - Async, loads config from AWS environment variables
  - DynamoDb::with_sdk_config(config) - Synchronous, uses custom SdkConfig
This gives clients full flexibility of the AWS SDK while maintaining backward compatibility. "dynamodb://" will continue to work when creating a Driver via Connect::new(url)

### Testing:
Tests will need AWS credentials environment variables set (even for DynamoDB Local):
  - AWS_ENDPOINT_URL_DYNAMODB=http://localhost:8000
  - AWS_REGION=us-east-1
  - AWS_ACCESS_KEY_ID=test
  - AWS_SECRET_ACCESS_KEY=test

  CI workflows will be updated accordingly.

asiefker:
Removed testing configuration from toasty-driver-dynamodb. The DDB
    driver can be customized either through env vars (see the
    workflow/ci.yml for examples), or a client can build the DDB client
    themselve, or pass a SdkConfig to the DynamoDb::with_sdk_config.

    tests/tests/dynamodb.rs was modified to set default AWS env vars if the
    values are not setup. For example, the DDB tests can be pointed at real
    DDB with some command line env vars.

    Fixes #388

### DIFF
--- a/tests/tests/mysql.rs
+++ b/tests/tests/mysql.rs
@@ -1,8 +1,8 @@
 #![cfg(feature = "mysql")]
 
 use mysql_async::prelude::Queryable;
-use tokio::sync::OnceCell;
 use toasty_driver_mysql::MySQL;
+use tokio::sync::OnceCell;
 
 struct MySqlSetup {
     pool: OnceCell<mysql_async::Pool>,
--- a/tests/tests/postgresql.rs
+++ b/tests/tests/postgresql.rs
@@ -1,9 +1,9 @@
 #![cfg(feature = "postgresql")]
 
 use std::sync::Arc;
+use toasty_driver_postgresql::PostgreSQL;
 use tokio::sync::OnceCell;
 use tokio_postgres::NoTls;
-use toasty_driver_postgresql::PostgreSQL;
 
 struct PostgreSqlSetup {
     client: OnceCell<Arc<tokio_postgres::Client>>,


crates/toasty/src/db/connect.rs
```diff
@@ -39,7 +39,22 @@ impl Connect {
...
+                        tokio::runtime::Runtime::new()
+                            .expect("Failed to create tokio runtime")
+                            .block_on(toasty_driver_dynamodb::DynamoDb::from_env(url))
+                    }
+                })
+                .join()
+                .expect("Failed to join driver initialization thread")
+                .map_err(toasty_core::Error::driver_operation_failed)?;
+                Box::new(driver)
+            }
```
carllerche, line 48: I'm not following why a new thread is spawned here.

tests/tests/dynamodb.rs
```diff
@@ -1,46 +1,61 @@
...
-                    .endpoint_url("http://localhost:8000")
-                    .load()
-                    .await;
-
-                Client::new(&config)
+    fn get_client(&self) -> &Client {
+        self.client.get_or_init(|| {
+            // Set default AWS environment variables if not already set
+            if std::env::var("AWS_REGION").is_err() {
+                std::env::set_var("AWS_REGION", "us-east-1");
```
carllerche: set_env is deprecated as it is unsound. Is there a different way to set defaults?

carllerche (changes requested):

crates/toasty/src/db/connect.rs
asiefker: `Connect::new()` is a standard fn, `DynamoDb::from_env` is async because it calls `aws_config::defaults(BehaviorVersion::latest()).load()` which is async. 

What is the idiomatic way to do this? Making `Connect::new` async seemed like overkill. I can move the thread::spawn inside DynamoDb and make that function sync instead.

tests/tests/dynamodb.rs
asiefker: I think there's a way to use the config builders from the AWS SDK instead. I'll research it more.

asiefker: The new PR rev has moved off of set_var. ConfigLoader is updated with the overrides.

crates/toasty/src/db/connect.rs
carllerche: Ah, in this case, the two best options are either a) make DynamoDb::from_env called lazily b) make Connect::new async. I think just making Connect::new async is the right way here (unless I'm missing something) because there is clearly a reason to do so (a db driver needing async work).

tests/tests/dynamodb.rs
carllerche: I still see set_var. Not sure what you mean by "moved off of". Given that this is test infrastructure, I am not blocking on changing this. It seems like there should be a better way than use set_var, but I am ok moving forward. I will +1 the PR.

PR #496, opened 2026-03-18
feat: add `.count(&mut dyn Executor)`

pikaju:
Add a `.count()` method to queries, allowing users to efficiently count matching rows without loading full records.

DynamoDB has native support for count operations, so it makes sense to have dedicated support for it in Toasty rather than forcing users to load all records and count client-side. Other aggregations (sum, avg, etc.) need to be investigated separately.

### Usage

```rust
let count = User::find_by_name("John")
    .count(&mut db)
    .await?;
// count: i64
```

### What changed

- Added `Query::count()` which rewrites the query to return `COUNT(*)` as a single `i64` result
- Added `Load` impl for `i64` so count results can be deserialized
- Generated `.count()` method on query builders via codegen
- Engine level changes to support `COUNT(*)` queries

### DIFF
--- a/crates/toasty/src/engine/exec/query_pk.rs
+++ b/crates/toasty/src/engine/exec/query_pk.rs
@@ -33,6 +33,15 @@ pub(crate) struct QueryPk {
 
     /// When true, return only the count of matching rows.
     pub count_only: bool,
+
+    /// Maximum number of items to return.
+    pub limit: Option<i64>,
+
+    /// Sort key ordering direction.
+    pub order: Option<stmt::Direction>,
+
+    /// Cursor for resuming a paginated query.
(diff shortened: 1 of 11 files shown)


crates/toasty/src/engine/exec/query_pk.rs
```diff
@@ -42,32 +45,66 @@ impl Exec<'_> {
...
+                            pk_filter: f,
+                            filter: action.row_filter.clone(),
+                            count_only: false,
+                        }
+                        .into(),
+                    )
+                    .await?;
+
+                all_rows.extend(res.rows.into_value_stream().collect().await?);
+            }
```
carllerche: Lots of duplication here, I bet you can get claude to do better :)

PR #498, opened 2026-03-18
docs: dynamoDB pagination with opaque cursor tokens

carllerche:
This adds design documentation for fixing DynamoDB pagination in Toasty. The current pagination implementation breaks for DynamoDB because it tries to reconstruct `LastEvaluatedKey` from row data, which is impossible for GSI queries and doesn't account for DynamoDB's evaluation-based `Limit` semantics.

[Rendered](https://github.com/tokio-rs/toasty/blob/claude/dynamodb-pagination-design-a5P2N/docs/dev/src/design/dynamodb-pagination.md)

### DIFF
--- a/docs/dev/src/design/dynamodb-pagination.md
+++ b/docs/dev/src/design/dynamodb-pagination.md
@@ -1,0 +1,166 @@
+# DynamoDB Pagination
+
+## Problem
+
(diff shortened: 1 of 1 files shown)


carllerche:
Closed by #610

PR #547, opened 2026-03-24
refactor: pagination refactor

asiefker:
The primary objective was to move Pagination logic into the engine. This will support databases that cannot build cursors from the result items, such as DynamoDB. 

This change implements paginated queries on DynamoDB base tables that have compound key - partition & sort. 
What's missing: 
* all + limit - DDB requires index access instead of allowing a user to scan a partition key. 
* GSIs - There's an issue with how Toasty creates GSIs with a partition & sort key

SQL databases use the new Pagination implementation.


carllerche:
Closed by #610

PR #579, opened 2026-03-29
fix: wrap One and OptionOne query structs with correct single-result types

carllerche:
One now wraps Query<Model> instead of Query<List<Model>>, and OptionOne
wraps Query<Option<Model>> instead of Query<List<Model>>. The conversion
from list queries happens inside from_stmt via .one() and .first().

Also adds Query<Option<T>>::into_rows() as the inverse of first() to
convert back to Query<List<T>> when needed (e.g., for set_scope in create).

https://claude.ai/code/session_01S16hRm6bWrp2utsfvx1str


carllerche:
Stale. Pretty sure this was fixed elsewhere.

PR #613, opened 2026-04-01
feat: add typed parameter list to QuerySql and Insert operation

MavenRain:


carllerche:
Whoops. It seems like I actually hit the same issue that motivated this and redid the work myself 😬  sorry.  I did end up using a different strategy: type inference. The reason for this is it seemed necessary to support native postgresql enums. It will be part of #665. I am going to close this PR since it seems that #665 handles it? If it doesn't, please open an issue or ping me on discord to discuss further: https://discord.gg/tokio (#toasty channel).

PR #662, opened 2026-04-08
feat: add TLS support to PostgreSQL driver

carllerche:
## Summary
This PR adds support for custom TLS connectors to the PostgreSQL driver, allowing users to connect with encrypted connections using either `native-tls` or `rustls`.

## Key Changes
- Made `PostgreSQL` a generic struct with a `Tls` type parameter (defaults to `tokio_postgres::NoTls`)
- Added `PostgreSQL::with_tls()` method to create drivers with custom TLS connectors
- Kept `PostgreSQL::new()` as a convenience method for unencrypted connections
- Updated the `Driver` trait implementation to support any `MakeTlsConnect<Socket>` implementation
- Implemented custom `Debug` for `PostgreSQL<Tls>` to avoid exposing TLS connector details
- Updated documentation with examples for both `native-tls` and `rustls` connectors
- Modified connection logic to use the configured TLS connector instead of hardcoded `NoTls`

## Implementation Details
- The `Tls` type parameter is bounded by `MakeTlsConnect<Socket> + Clone + Send + Sync + 'static` with additional trait bounds on associated types for async compatibility
- Both `PostgreSQL::new()` and `PostgreSQL::with_tls()` now delegate to a common implementation in the generic impl block
- The TLS connector is cloned for each connection attempt, supporting the existing connection pooling and database creation workflows

https://claude.ai/code/session_01LT238p2ssNBs8i2aMubYcU

### DIFF
--- a/crates/toasty-driver-postgresql/src/lib.rs
+++ b/crates/toasty-driver-postgresql/src/lib.rs
@@ -5,10 +5,45 @@
 //!
 //! # Examples
 //!
+//! Connecting without TLS:
+//!
 //! ```no_run
 //! use toasty_driver_postgresql::PostgreSQL;
 //!
 //! let driver = PostgreSQL::new("postgresql://localhost/mydb").unwrap();
+//! ```
+//!
+//! Connecting with TLS using [`native-tls`](https://docs.rs/native-tls) (requires
+//! the `native-tls` feature):
+//!
+//! ```no_run,ignore
+//! use toasty_driver_postgresql::PostgreSQL;
+//! use postgres_native_tls::MakeTlsConnector;
+//! use native_tls::TlsConnector;
+//!
+//! let connector = TlsConnector::builder()
+//!     .danger_accept_invalid_certs(true) // for development only
+//!     .build()
+//!     .unwrap();
+//! let tls = MakeTlsConnector::new(connector);
+//!
+//! let driver = PostgreSQL::with_tls("postgresql://localhost/mydb", tls).unwrap();
+//! ```
+//!
+//! Connecting with TLS using [`rustls`](https://docs.rs/rustls) (requires the
+//! `rustls` feature):
+//!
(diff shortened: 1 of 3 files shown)


crates/toasty-driver-postgresql/src/lib.rs
```diff
@@ -35,22 +68,73 @@ use crate::{statement_cache::StatementCache, r#type::TypeExt};
...
+    ///
+    /// let connector = TlsConnector::builder()
+    ///     .danger_accept_invalid_certs(true) // for development only
+    ///     .build()
+    ///     .unwrap();
+    /// let tls = MakeTlsConnector::new(connector);
+    ///
+    /// let driver = PostgreSQL::with_tls("postgresql://rds-host/mydb", tls).unwrap();
+    /// ```
+    pub fn with_tls(url: impl Into<String>, tls: Tls) -> Result<Self> {
```
brandonpike, line 137: Many users will be used to passing ssl in through the url `"sslrootcert=<.pem> sslmode=verify-full"`.

brandonpike (approved):

crates/toasty-driver-postgresql/src/lib.rs
carllerche: Is that how it is actually done? You have a link to a rust crate that does this?

brandonpike: sqlx example:

- [url parse](https://github.com/launchbadge/sqlx/blob/18ffed2af676cbb7ae2f098282fd2465566b02b1/sqlx-postgres/src/options/parse.rs#L52-L58)
- [tls_native_tls handshake](https://github.com/launchbadge/sqlx/blob/main/sqlx-core/src/net/tls/tls_native_tls.rs#L42-L62)

carllerche:
Closing in favor of #663

PR #665, opened 2026-04-09
feat: add native database enum type support for embedded enums

carllerche:
String-label embedded enums now use the best available enum
representation for the target database by default:
- PostgreSQL: CREATE TYPE ... AS ENUM (named type)
- MySQL: inline ENUM('a', 'b', ...) column type
- SQLite: TEXT column with CHECK constraint
- DynamoDB: plain string attribute (no change)

Key changes:
- Add db::Type::Enum variant with name and labels
- Default string labels now use snake_case (e.g., InProgress -> "in_progress")
- Parse #[column(type = enum)] and #[column(type = enum("name"))] attributes
- Parse #[column(type = text)] to opt out of native enum storage
- Validate label lengths (max 63 bytes for PostgreSQL NAMEDATALEN)
- SQL serialization adapts per database flavor
- PostgreSQL push_schema emits CREATE TYPE before CREATE TABLE
- SQLite column definitions include CHECK constraints

https://claude.ai/code/session_01KdZ4TJjyiueRJcizjf9zMo


crates/toasty-sql/src/stmt/column_def.rs
```diff
@@ -14,15 +17,48 @@ pub struct ColumnDef {
...
-    pub(crate) fn from_schema(column: &Column, _storage_types: &driver::StorageTypes) -> Self {
+    pub(crate) fn from_schema(
+        column: &Column,
+        _storage_types: &driver::StorageTypes,
+        capability: &Capability,
+    ) -> Self {
+        // For SQLite enum columns: store as TEXT with a CHECK constraint instead
+        // of db::Type::Enum. The CHECK constraint restricts values to the
+        // declared variants.
+        if let db::Type::Enum(type_enum) = &column.storage_ty
```
carllerche, line 33: The structure of this method is strange.

PR #668, opened 2026-04-09
feat: error on limit() for NoSQL drivers, simplify DynamoDB query_pk

asiefker:
## Summary

- **`feat`**: Return a clear `UnsupportedFeature` error when `query.limit(n)` is used on a NoSQL driver. DynamoDB's `Limit` parameter caps *evaluated* items (not returned items), so a filter expression can silently produce fewer results than requested — or zero, even when matching rows exist. The 1 MB per-request cap compounds this regardless of filter presence. Users should use `paginate()` instead, which maps naturally to DynamoDB's per-request semantics.
- **`refactor`**: Eliminate duplicated `QueryFluentBuilder` setup in `exec_query_pk` — the primary key path and GSI path shared identical builder code. Now a single builder is constructed and `.index_name()` attached conditionally.

## Changes

- `crates/toasty/src/engine/plan/statement.rs` — guard in `plan_data_loading_nosql()` returns `unsupported_feature` for any `Limit::Offset` query on a NoSQL driver (mirrors the existing `count()` check)
- `crates/toasty-driver-dynamodb/src/op/query_pk.rs` — consolidate duplicated query builder
- `crates/toasty-driver-integration-suite/src/tests/crud_sort_limit.rs` — add `limit_not_supported_on_nosql` test (`requires(not(sql))`)
- `crates/toasty-driver-integration-suite/src/tests/crud_composite_key_pagination.rs` — mark `limit_composite_key` as `requires(sql)` since it uses `.limit()`

## Test plan

- [ ] `cargo test` passes (SQLite, no external services)
- [ ] `cargo test -p tests --features dynamodb` exercises the new error path on a live DynamoDB instance
- [ ] Existing `limit_offset` and `paginate` tests continue to pass on SQL drivers unchanged

### DIFF
--- a/crates/toasty-driver-dynamodb/src/op/query_pk.rs
+++ b/crates/toasty-driver-dynamodb/src/op/query_pk.rs
@@ -25,74 +25,45 @@ impl Connection {
             .as_ref()
             .map(|expr| ddb_expression(&cx, &mut expr_attrs, false, expr));
 
-        // Build the query based on whether we're querying primary key or an index
-        let result = if let Some(index_id) = op.index {
-            let index = schema.db.index(index_id);
+        // Build base query, then conditionally attach an index
+        let mut query = self
+            .client
+            .query()
+            .table_name(&table.name)
+            .key_condition_expression(key_expression)
+            .set_filter_expression(filter_expression)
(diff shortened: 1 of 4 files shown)


asiefker:
We opted to implement this functionality instead of returning an Error. 

See #674

PR #671, opened 2026-04-10
fix: percent encoding in postgres connection strings

### LINKED ISSUE #661: toasty-driver-postgresql: Unable to use some passwords due to percent-encoding
The postgresql driver (and maybe others) are using the `url` crate to parse the connection configuration. The url crate percent-encodes everything.
But tokio-postgres::Config doesn't decode that and assumes it raw, making it **impossible** to connect to databases with a variety of passwords with symbols in them.

Maybe the best idea would be to simply do `let config: Config = url.into()`, `try_into()` or similar and throw out most of the code dealing with the URL. This then would allow configuring a plethora of things on its own as documented in https://docs.rs/tokio-postgres/latest/tokio_postgres/config/struct.Config.html#keys

That `From` impl in tokio-postgres uses its own UrlParser, and also `percent-encoding` create to decode the password; that crate in turn is strict and just leaves it raw if the next two characters after % don't make up a valid percent-encoded character, so often the password can be raw as well even if it includes `%` in it.

I'm not sure why the URL parsing is hand-coded in `toasty-driver-postgresql` instead of making use of what tokio-postgres provides for just parsing an URL string into the config, but if the hand-coded thing has to remain, we need to percent_decode various things to support connecting to databases with various symbol containing passwords. If switching to tokio-postgres provided parsing, then that might need some documentation updates to point towards all the extra config via the URL being supported for advanced use cases.

conradludgate:
## Summary

Closes #661

`url::Url::username()`, `password()`, and `path()` return percent-encoded strings. We were passing these directly to `tokio_postgres::Config`, which expects decoded values. This made it impossible to connect with credentials or database names containing characters like `@`, `#`, `/`, etc.

This matches libpq's behavior, which percent-decodes all URI components via [`conninfo_uri_decode`](https://github.com/postgres/postgres/blob/REL_17_STABLE/src/interfaces/libpq/fe-connect.c#L6764-L6821). libpq calls this (via [`conninfo_storeval`](https://github.com/postgres/postgres/blob/REL_17_STABLE/src/interfaces/libpq/fe-connect.c#L6913-L6916)) for:
- [user (L6454)](https://github.com/postgres/postgres/blob/REL_17_STABLE/src/interfaces/libpq/fe-connect.c#L6454)
- [password (L6467)](https://github.com/postgres/postgres/blob/REL_17_STABLE/src/interfaces/libpq/fe-connect.c#L6467)
- [dbname (L6599)](https://github.com/postgres/postgres/blob/REL_17_STABLE/src/interfaces/libpq/fe-connect.c#L6599)

## Changes

- Decode username, password, and dbname with `percent_encoding::percent_decode_str` before passing to `Config`
- Username and dbname return an error on invalid UTF-8; password is passed as raw bytes
- Added `percent-encoding` as a direct dependency (already a transitive dep via `url` and `tokio-postgres`)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -87,6 +87,7 @@ mysql_async = { version = "0.36.2", default-features = false, features = [
 ] }
 pulldown-cmark = { version = "0.13", default-features = false }
 pluralizer = "0.5.0"
+percent-encoding = "2.3.2"
 postgres-types = { version = "0.2.13", features = ["with-uuid-1", "array-impls"] }
 pretty_assertions = "1.4.1"
 proc-macro2 = "1.0.106"
--- a/crates/toasty-driver-postgresql/Cargo.toml
+++ b/crates/toasty-driver-postgresql/Cargo.toml
@@ -38,6 +38,7 @@ bigdecimal = { workspace = true, optional = true }
 jiff = { workspace = true, optional = true }
 lru.workspace = true
 tracing.workspace = true
+percent-encoding.workspace = true
 postgres-types.workspace = true
 tokio.workspace = true
 tokio-postgres.workspace = true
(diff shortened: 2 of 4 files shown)


carllerche (approved):
Thanks :+1:

PR #676, opened 2026-04-12
feat: add json/jsonb for psql

cougarwww:
update  #[serialize(json/jsonb)] to use postgres JSON/JSONB backend


carllerche:

crates/toasty-core/src/driver/capability.rs
```diff
@@ -76,6 +76,14 @@ pub struct Capability {
...
     pub bigdecimal_implemented: bool,
 
+    /// Whether the database supports JSONB (binary JSON) column type.
+    /// PostgreSQL: true. SQLite/MySQL/DynamoDB: false (will fallback to TEXT).
+    pub supports_jsonb: bool,
+
+    /// Whether the database supports JSON column type.
+    /// MySQL/PostgreSQL: true. SQLite/DynamoDB: false (will fallback to TEXT).
+    pub supports_json: bool,
+
```
pikaju, line 86: It would be nice to call these `native_json(b)` and put them underneath the other `native_...` bools. The explanation per database would (in my opinion) be better in the instantiations for that specific DB.

crates/toasty-core/src/schema/app/field/primitive.rs
```diff
@@ -17,6 +17,10 @@ use crate::{schema::db, stmt};
 pub enum SerializeFormat {
     /// Serialize the value as JSON using `serde_json`.
     Json,
+    /// Serialize the value as JSONB (binary JSON) using `serde_json`.
+    /// JSONB is a PostgreSQL-specific binary format that supports indexing
+    /// and more efficient operations.
+    Jsonb,
```
pikaju, line 23: I'm not sure this is the correct place to put this, I believe this enum is specifically meant for the _serde_ format (e.g. JSON, YAML, TOML, etc) and not how it is stored on disk. I think the `jsonb` information should be defined by the `#[column(type = jsonb)]` attribute. In Postgres, type `jsonb` could be the default for all `#[serialize(json)]` columns.

crates/toasty-core/src/schema/db/ty.rs
```diff
@@ -112,9 +112,27 @@ impl Type {
...
+                        }
+                        return Ok(Type::Text);
+                    }
+                    if custom == "JSON" && !capability.supports_json {
+                        // Fallback to TEXT for databases without JSON support
+                        return Ok(Type::Text);
+                    }
+                }
+                Ok(ty.clone())
+            }
```
pikaju, line 135: Instead of using `Type::Custom` I think `Json` and `Jsonb` deserve dedicated database-level `Type` variants

crates/toasty-core/src/stmt/ty.rs
```diff
@@ -332,6 +352,10 @@ impl Type {
         Ok(match (value, self) {
             // Identity
             (value @ Value::String(_), Self::String) => value,
+            // String is compatible with Json (identity cast)
+            (value @ Value::String(_), Self::Json) => value,
+            // String is compatible with Jsonb (identity cast)
+            (value @ Value::String(_), Self::Jsonb) => value,
```
pikaju, line 358: I don't understand, why doesn't this cast convert the type to `Value::Json`? It seems to be leaving it as `Value::String` after the cast

crates/toasty-core/src/stmt/ty.rs
```diff
@@ -408,6 +432,10 @@ impl Type {
             // Simple types must match exactly
             (Type::Bool, Type::Bool) => true,
             (Type::String, Type::String) => true,
+            // A JSON-encoded string is assignable to a Json type
+            (Type::String, Type::Json) => true,
+            // A JSONB-encoded string is assignable to a Jsonb type
+            (Type::String, Type::Jsonb) => true,
```
pikaju, line 438: What is the reasoning here?

crates/toasty-macros/src/model/expand/schema.rs
```diff
@@ -113,13 +113,27 @@ impl Expand<'_> {
...
+                            let (stmt_ty, storage_ty) = match format {
+                                SerializeFormat::Json => (
+                                    quote!(#toasty::core::stmt::Type::Json),
+                                    quote!(Some(#toasty::core::schema::db::Type::Custom("JSON".to_string()))),
+                                ),
+                                SerializeFormat::Jsonb => (
+                                    quote!(#toasty::core::stmt::Type::Jsonb),
+                                    quote!(Some(#toasty::core::schema::db::Type::Custom("JSONB".to_string()))),
+                                ),
+                            };
```
pikaju, line 131: As mentioned above, I think this deserves a dedicated type

docs/guide/src/field-options.md
```diff
@@ -70,19 +70,19 @@ struct User {
...
+| `int`, `i8`, `i16`, `i32`, `i64`  | Integer (various sizes)                   |
+| `uint`, `u8`, `u16`, `u32`, `u64` | Unsigned integer                          |
+| `text`                            | Text                                      |
+| `varchar(N)`                      | Variable-length string with max length    |
+| `numeric`, `numeric(P, S)`        | Decimal with optional precision and scale |
+| `binary(N)`, `blob`               | Binary data                               |
+| `timestamp(P)`                    | Timestamp with precision                  |
+| `date`                            | Date                                      |
+| `time(P)`                         | Time with precision                       |
+| `datetime(P)`                     | Date and time with precision              |
```
pikaju, line 85: There are a bunch of these formatting changes in this PR, in my opinion they should be extracted into a separate PR because I would argue they are not objectively better

docs/guide/src/field-options.md
```diff
@@ -397,15 +437,12 @@ that support `varchar`.
 #     #[auto]
 #     id: u64,
 #     title: String,
-#     #[serialize(json)]
-#     tags: Vec<String>,
```
pikaju, line 401: Why remove this field?

examples/hi-jsonb/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
```
pikaju, line 1: I don't think JSONB deserves its own example project, otherwise we would end up having one example project per toasty type

Cargo.toml
```diff
@@ -106,13 +110,19 @@ proptest = "1"
...
 const-oid = { version = "0.9.6", default-features = false, features = ["db"] }
 sha2 = { version = "0.11", default-features = false }
 x509-cert = { version = "0.2.5", default-features = false, features = ["std"] }
 rustls-platform-verifier = "0.6"
 rustls-pemfile = "2"
-rustls-webpki = { version = "0.103", default-features = false, features = ["alloc", "std"] }
+rustls-webpki = { version = "0.103", default-features = false, features = [
+    "alloc",
+    "std",
+] }
```
pikaju, line 128: See above, the formatting should be discussed in a separate PR in my opinion.

crates/toasty-core/src/stmt/ty.rs
pikaju: After some thinking it seems there is no new `stmt::Value` variant added in this PR. However as far as I know the `stmt::Type` and `stmt::Value` enums should be tightly linked.

I believe a better approach for now would be to leave the `stmt::Value` and `stmt::Type` as it is for now and just modify `db::Type` to support `Json` and `Jsonb`. Then the default `from_app` response would be `Jsonb` for Postgres when using a `#[serialize(json)]` column.

pikaju (changes requested):
Thank you for building this, I believe we need to make some changes to the approach though before this can get merged. @carllerche

carllerche:
Thanks for the work. This doesn't line up with the direction I think we need to take. jsonb is on the roadmap but my guess is it will go a different path.

PR #680, opened 2026-04-14
feat: add application_name to Postgresql client

brandonpike:
Closes #679

---

Should be noncontroversial.

PR #681, opened 2026-04-15
fix: avoid reusing broken connections

### LINKED ISSUE #678: Toasty connection pool doesn't recover from a database restart
If say an axum backend is running with toasty::Db in its state and the database restarts, all tries to do any database operations appear to simply error with `connection error: db error` instead of reconnecting.
This means potential longer production downtime just because of a split second database restart.

brandonpike:
## Summary

Closes #678

When connections break, they were being returned to the connection pool to be reused. This commit adds an is_valid method to the Connection trait that we will use to exit the connection join handle. The connection pool will then check the join handle during recycling to see if it can be put back in the pool.

## Notes

Each driver will have to decide what `is_valid` means for their Connections.


tests/tests/postgresql.rs
```diff
@@ -137,3 +139,66 @@ async fn url_encoding() {
         .await
         .expect("failed to drop test role");
 }
+
+#[tokio::test]
+async fn pool_recovers_db_crash() {
```
brandonpike, line 198: since each DB has to implement `is_valid`, maybe every DB should have a test validating this.

tests/tests/postgresql.rs
```diff
@@ -137,3 +139,66 @@ async fn url_encoding() {
...
+        .query_pairs_mut()
+        .append_pair("application_name", app_name);
+    let driver = PostgreSQL::new(tagged_url.as_str()).expect("driver creation failed");
+    let mut db = Db::builder()
+        .models(toasty::models!(User))
+        .build(driver)
+        .await
+        .expect("Db build failed");
+
+    db.push_schema().await.unwrap();
```
brandonpike, line 236: This test actually leaks the schema since it never gets dropped.

Surely I can make this idempotent with RAII and not have to dirty my tests with `DROP ... IF EXISTS`

brandonpike:
Note: the Postgres test depends on the merge of #680 - we're using the application_name of the client to ensure we kill the right backend (and only the right backend) in order to simulate a DB or connection crash.

tests/tests/postgresql.rs
```diff
@@ -137,3 +139,125 @@ async fn url_encoding() {
...
+        .await
+        .expect("pg_terminate_backend failed");
+
+    tokio::time::sleep(Duration::from_millis(500)).await;
+
+    // The connection gets closed when the backend gets terminated
+    assert!(User::filter_by_id(1)
+        .exec(&mut db)
+        .await
+        .is_err_and(|err| err.to_string().eq("connection closed")));
```
leio: Maybe I'm not fully following it right, but if it gives an error for each connection in the pool that died, and the application code needs to retry to get the result, then that's not something that will make for maintainable app code.
In my opinion it just needs to find a working connection if possible and succeed. It's OK to error if the database is still down, of course.

carllerche:

tests/tests/postgresql.rs
leio: While I don't think it's nice to feel this in the application code in many cases, it's of course preferable to at least recover with some backend calls erroring out until the pool has only healthy connections again, rather than all of them remaining broken indefinitely - until we get the nicer things figured out in design and implementation

carllerche:
Whats the latest re: this PR? It is still flagged as draft, so I assume it isn't ready yet.

carllerche:
Handled in another PR.

PR #684, opened 2026-04-15
feat: enable timestamp migration prefix

heyjul:
Closes #652
It still works if the user changes prefix style between migrations but from Timestamp to Sequential, the generated name will be :
- 20260415_143200_migration.sql [Timestamp]
- 20260416_migration.sql [Sequential after Timestamp]


carllerche:

pikaju (approved):

PR #685, opened 2026-04-15
feat: allow multiple external crate glob inputs in models! macro

philip-hartmann:
The models! macro now accepts multiple inputs such as `some_crate::*` and registers every discoverable model from the specified crates.

Previously, only types could be used after the first input because `$rest:ty` does not support glob imports like `some_crate::*`. This change extends the macro to support such patterns.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,8 @@ members = [
 
     # Tests
     "tests",
+    "tests/tests/fixtures/post",
+    "tests/tests/fixtures/user",
 
     # Benchmarks
     "benches",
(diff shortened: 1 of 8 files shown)


joeydewaal (dismissed):
Could you add a test for this in `tests/tests/discover.rs`? Otherwise looks good!

carllerche:

philip-hartmann:
I wrote a couple test cases which should cover the use cases of the `models!` macro and put them in a new file `tests/tests/models_macro.rs`. I also added two small crates in `tests/tests/fixtures/` to use them as test inputs.

pikaju (approved):
Thank you!

pikaju:
@philip-hartmann could you run `cargo fmt` and merge main so the pipeline passes?

PR #687, opened 2026-04-16
feat: add support for floats

joeydewaal:
This pr adds support for floating points, both `f32` and `f64`.


carllerche:
Thanks for the PR and sorry for the delay.

I actually think we probably need to **remove** Hash and Eq from Value. I am going to (try) to explore that. It is there to support storing Value in HashMap, but that can probably be done a different way.


I'm pretty swamped with [TokioConf](https://tokioconf.com) (next week!!) but will be circling back to review everything afterwards.

crates/toasty-core/src/stmt/value.rs
```diff
@@ -405,6 +414,89 @@ impl AsRef<Self> for Value {
...
+            (Value::I16(a), Value::I16(b)) => a == b,
+            (Value::I32(a), Value::I32(b)) => a == b,
+            (Value::I64(a), Value::I64(b)) => a == b,
+            (Value::U8(a), Value::U8(b)) => a == b,
+            (Value::U16(a), Value::U16(b)) => a == b,
+            (Value::U32(a), Value::U32(b)) => a == b,
+            (Value::U64(a), Value::U64(b)) => a == b,
+            // Use bit-level equality for floats so NaN == NaN and the impl is consistent with Hash
+            (Value::F32(a), Value::F32(b)) => a.to_bits() == b.to_bits(),
+            (Value::F64(a), Value::F64(b)) => a.to_bits() == b.to_bits(),
```
pikaju: I agree with Carl, removing the `Eq` derive is probably better, what's the use case for using `Value` as a hash map key?

pikaju (dismissed):

carllerche:
I added a PR to remove Eq / Hash on Value: #705

joeydewaal:
The manual PartialEq and Hash impl are now removed.

carllerche (approved):
LGTM 👍 thanks

PR #689, opened 2026-04-16
test: restrict tx rollback tests to auto_increment sql drivers

leruaa:
This PR:

* Removes `#[auto]` from `u64` key fields in `relation_has_many_boxed_fk` test and supplies explicit IDs, making them compatible with drivers that don't support auto-increment
* Restricts `create_with_has_many_rolls_back_on_failure` and `create_with_has_one_rolls_back_on_failure` to `auto_increment` drivers, since the explicit transaction wrapping they verify only occurs with auto-increment IDs

I needed this to fix tests failures with the DuckDB driver I'm building and don't support `auto_increment`.


carllerche:

pikaju (approved):
I think this is fine

PR #674, opened 2026-04-10
feat: add Limit::Offset support to DynamoDB driver

asiefker:
## Summary

- Introduces `QueryPkLimit` enum in `toasty-core` (`Cursor` and `Offset` variants), replacing the separate `limit`/`cursor` fields on `operation::QueryPk` and `mir::QueryPk`
- Plumbs the previously-dropped `offset` field through `extract_query_pk_pagination` in the plan phase
- DynamoDB driver `exec_query_pk` uses `into_paginator()` to loop until `limit + offset` items are collected, discards the first `offset` items in-stream (never storing them), and returns `next_cursor: None` for offset queries
- Fixes offset extraction bug: `Value::from(usize)` produces `Value::U64` but extraction previously only matched `Value::I64`

## Test plan

- [ ] Un-gated composite-key limit tests now run on DynamoDB
- [ ] New `limit_offset_composite_key` test: base table `QueryPk` path with `.limit()`, `.limit().offset()`, and limit > result set
- [ ] New `limit_offset_gsi` test: GSI `QueryPk` path with same coverage
- [ ] New `dynamodb_limit_boundary` tests: 4 DDB-only tests with 10 KB item payloads forcing multiple Query API calls, covering base table and GSI paths for both limit-only and limit+offset
- [ ] `cargo test` (SQLite) green
- [ ] `cargo test --features dynamodb` green

> **Note:** This PR depends on #664. Once that merges the diff here will shrink to just the DynamoDB offset changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #694, opened 2026-04-17
feat: add #[version] optimistic concurrency control for DynamoDB

asiefker:
## Summary

> **Stacked on #674** — this PR depends on the `ddb-limit-support` branch. Once #674 merges, the diff here shrinks to just the `#[version]` changes.

- Adds a `#[version]` field attribute that enables optimistic concurrency control (OCC) on DynamoDB models
- The version field (must be `u64`) is auto-initialized to `1` on create, atomically incremented on instance update, and checked on instance delete
- Stale writes — where another writer advanced the version since the record was loaded — return an error rather than silently overwriting
- Supports all three DynamoDB update paths: single-key `UpdateItem`, multi-key `TransactWriteItems`, and unique-index surgery via `TransactWriteItems`
- Delete path extended to carry and evaluate version conditions through the full stack (`stmt::Delete.condition` → lower → MIR → exec → `operation::DeleteByKey` → DDB driver)
- Integration tests in `field_version.rs`: create initializes version, instance update increments version, stale update fails, duplicate create fails, batch insert version check, query-based multi-key update works, unique-index update increments version, unique-index stale update fails, delete checks version, stale delete fails
- Docs added to `field-options.md`, `updating-records.md`, `deleting-records.md`

## Test plan

- [ ] `cargo test` (SQLite) green
- [ ] `cargo test --features dynamodb --no-default-features -- --test-threads=1` green — all 10 `field_version` integration tests pass


carllerche:
Ok, I tried to update the PR to be based on main (with the other PRs merged).

crates/toasty-macros/src/model/expand/model.rs
```diff
@@ -226,10 +229,48 @@ impl Expand<'_> {
...
+        let Some(version_index) = self.model.kind.as_root().and_then(|r| r.version_field) else {
+            return quote! { __delete };
+        };
+
+        let field = &self.model.fields[version_index];
+        let index_tokenized = util::int(version_index);
+        let field_ident = &field.name.ident;
+
+        quote! {
+            __delete.set_condition(
```
carllerche, line 259: I wonder if this should be set in the engine instead of in the macro.

asiefker: My thinking was to keep the version logic centralized in the expansion, then the rest of the code base can treat it as condition and not worry about it. 

I'm not opposed to moving this though. Why do you think the engine would be a better place?

carllerche: The question for me is whether or not it would be *valid* to explicitly set the version. E.g. one can submit a manual stmt that sets version or does something wonky. Should that be permitted?

asiefker: I've been using DDB for many years, usually with the DynamoDbEnhancedClient. It provides a version annotation to mange OCC for you. I don't recall ever needing to set it to a value. In my experience, the bigger risk by far, is needing to do an out of band backfill of some sort. Those tools are frequently written in Python, and may forget to update the version, running the risk of corrupting in flight work.

carllerche: My guess is if we need to do anything more complicated, then having the engine know about it would be good. That said, it can be a future enhancement and what you have is good enough to get this merged. I will probably track it as an open question in an issue.

carllerche (approved):
:+1: thanks

PR #703, opened 2026-04-18
fix: gate simplification rules on expression stability

### LINKED ISSUE #236: Simplification rules assumes that if two expressions are syntactically identical, they are equivalent.
See [this comment](https://github.com/tokio-rs/toasty/pull/217#issuecomment-3613278138)

Generally, I think this exposes a bigger issue with the engine setup. Right now, it simplifies the AST, which has limitations. My guess is we will want to move simplification into the AST -> HIR process. This way, we can split out sub-statements to prevent having simplification recurse into them and hopefully avoid running the simplification pass many times.

MavenRain:
equivalence was unsound for non-deterministic expressions such as `LAST_INSERT_ID()`, which may yield different values on each evaluation.  Introduces `Expr::is_equivalent_to`, which combines syntactic identity with `is_stable`, and route idempotent, absorption, complement, factoring, range-to-equality, contradicting-equality, variant-tautology, OR-to-IN, and `Project == Project` self-compare rewrites through it.  Regression tests exercise each rule with `LAST_INSERT_ID()` to pin the new semantics.

Closes #236.

### DIFF
--- a/crates/toasty-core/src/stmt/expr.rs
+++ b/crates/toasty-core/src/stmt/expr.rs
@@ -275,6 +275,23 @@ impl Expr {
         }
     }
 
+    /// Returns `true` if `self` and `other` are syntactically identical **and**
+    /// both sides are stable.
+    ///
+    /// This is the soundness-preserving comparison used by simplification
+    /// rules that rewrite on the assumption that two equal sub-expressions
+    /// produce the same value (idempotent, absorption, complement,
+    /// range-to-equality, OR-to-IN, factoring, variant tautology).
+    ///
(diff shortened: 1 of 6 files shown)


pikaju (commented):
@carllerche I think this is a good change, but my two cents: Perhaps we should modify the implementation of `PartialEq` instead so this mechanism is always used? I think it ties nicely into Rusts `PartialEq` vs. `Eq` semantics

carllerche:
I suspect we will eventually get to implementing equivalence classes. Before we get there, we are most likely going to have to restructure the engine to move simplification after lowering.

Issue #326, opened 2026-02-18
Primitive trait has evolved to be a "Field" trait


carllerche:
Though, it doesn't handle relation fields...

Also the `UpdateBuilder` associated type could be renamed to something shorter.

carllerche:
The rename and the type cleanup both landed:

- `Primitive` is now `Field` (`crates/toasty/src/schema/field.rs:11`).
- `UpdateBuilder` is now `Update<'a>`.

`Field` is implemented for primitives, `Vec<u8>`, `Option<T>`, `Arc<T>` / `Rc<T>` / `Box<T>`, jiff, and numeric types. Relation types (`BelongsTo`, `HasMany`, `HasOne`) still don't implement `Field`, but that's a separate design question — `Field` describes storage-bearing fields, and folding relations into the same trait needs its own discussion. Open a focused issue if you want to pursue it.

PR #706, opened 2026-04-18
refactor: handle relation fields in Relation trait

MavenRain:
Move the inline FieldTy::BelongsTo / HasMany / HasOne construction from the #[derive(Model)] macro into overridable methods on the Relation trait. The three methods (`belongs_to_field_ty`, `has_many_field_ty`, `has_one_field_ty`) default to panics so that applying a relation attribute to a non-wrapper type fails loudly.

Addresses the "doesn't handle relation fields" portion of #326; the Primitive -> Field rename and UpdateBuilder -> Update rename were already landed in #568.


carllerche:
Thanks. I approve of moving more logic out of the macro.

PR #707, opened 2026-04-19
feat: latest_by query

hlsxx:
### Summary

This PR adds a wrapper query method `latest_by`, which provides a shortcut for ordering the latest records by a specific column.

With order_by:

```rust
User::all()
  .order_by(User::fields().created_at().desc())
  .first()
```

With latest_by:

```rust
User::all()
  .latest_by(User::fields().created_at())
  .first()
```

### Future work

Restrict `latest_by()` to only accept fields with reasonable orderable types such as timestamps, auto-increment ids, and  uuids, preventing nonsensical ordering by fields like name or email at compile time.


carllerche:
Thanks. If this is a shortcut for `order_by` it probably should return the query (like order_by does) instead of a new type. What do you think?

hlsxx:
Maybe it's a good idea to keep them distinguished because of the potential for specific field type restrictions?

The `From<Path<T, U>> for LatestBy` impl currently accepts any field type, but for example an added trait `LatestByField` marker trait could be added in the future to restrict `latest_by()` to only orderable types like auto-increment keys, jiff::Timestamp etc., giving compile-time safety.

However, I'm not sure if this fits the current design — there might be cleaner approaches. Happy to hear your thoughts :)

carllerche:
Sorry, I misread the PR initially. It seems that the idea is you have `fn latest_by` take `impl Into<LatestBy>`. If I'm reading this, *all* paths implement Into<LatestBy>? If so, I would probably just have `latest_by` take a path (well, impl Into path)... not 100% sure. Maybe you tried that and it didn't line up?

The other direction I could see is not all paths impl Into<LatestBy> only paths that point to "time" based types.

Regardless, LatestBy probably shouldn't be in `toasty_core` as it isn't part of the core stmt AST. It should be in `toasty`.

hlsxx:
Thank you for your observation :) I removed the `LatestBy` type and rethought the approach a bit. I introduced a `Recency` trait which currently restricts `fn latest_by` to time-based field types from the jiff library.
 
This also leaves room for implementing the opposite method `oldest_by` in the future, as well as potentially supporting auto-increment fields via something like `impl Recency for AutoIncrement<u64>` once `#[auto]` fields have a dedicated wrapper type, sure if this feature passed.

carllerche:
I've been debating this a bunch. I don't think the trait is buying much in terms of preventing bugs. E.g. there are other types of fields that would make sense to order by recency, like a version field. Also, latest_by as a phrase only makes sense when the "time" field is only in the past (vs future), which the trait does not guard against.

So, I opted to just remove the trait and make it a conventional helper. Feel free to open an issue or follow up PR if you disagree.

PR #736, opened 2026-04-20
chore: create transaction from immutable `Db`

joeydewaal:
## Summary

This pr changes the signature of `transaction` and `transaction_builder` on `Db` to `&self` (in stead of `&mut self`). This was probably forgotten in #569. 

## Type of change

- [ ] Other (please explain):

## Checklist

- [x] `cargo test` passes

## Notes for reviewers

none


joeydewaal:
Ci should be unrelated

carllerche:
I believe that this is an intentional decision. Cc @pikaju. The intent is to prevent accidentally using the original Db handle when there is an open transaction. If there is too much API friction we could possibly consider a runtime check instead.

crates/toasty/src/db.rs
```diff
@@ -168,3 +168,22 @@ impl Executor for Db {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::Db;
+
+    #[test]
+    fn this_should_compile() {
+        #[allow(unused)]
+        async fn shared_ref(db: &Db) -> crate::Result<()> {
```
pikaju, line 179: Does this test test what you want it to? Since you drop the tx handles before starting a new tx this would've worked before, right? Shouldn't you drop them at the end? Or am I missing something?

pikaju (changes requested):

crates/toasty/src/db.rs
joeydewaal: This test only checks if the `shared_ref` function compiles. But yeah, the `std::mem::drop` calls might be a bit weird. I'll remove them.

joeydewaal:
Closing this since the current api avoids a common pitfall where one can accidentally use a `Db` while there is an active `Transaction` open that should probably be used instead.

carllerche:
Ref: #752. Adding some docs explaining this.

PR #745, opened 2026-04-23
feat: add begins_with and LIKE string prefix filter operators

asiefker:
## Summary

Adds two new string prefix filter operators to Toasty:

**`begins_with(prefix)`** — available on `Path<T, String>` fields. On DynamoDB this maps to the native `begins_with(attr, :val)` function,  routed automatically to `KeyConditionExpression` (sort key) or `FilterExpression` (non-key attribute) by the existing planner split.
On SQL it lowers to `LIKE 'prefix%'`.

**`like(pattern)`** — available on `Path<T, String>` fields for SQL drivers. The caller supplies the full pattern including `%` / `_`  wildcards. 
Not supported by DynamoDB — panics with a clear message if   reached.

## Type of change

- [X] Small / obvious change — bug fix, docs, internal cleanup, or test
- [ ] Other (please explain):

## Checklist

- [x] `cargo test` passes


carllerche (approved):
LGTM thanks. It should be possible to translate compatible Like expressions -> begins_with. I tracked this as a follow up issue: https://github.com/tokio-rs/toasty/issues/763.

leio:
My feedback on this from a PostgreSQL angle:

1. As a person unfamiliar with DynamoDB, `starts_with` would sound much more natural to me than `begins_with`, because other ORMs exposing this tend to use that, and rust `String` itself has `starts_with`
2. It isn't exposed for `Option<String>` fields unexpectedly to me
3. It fails to escape special characters for PostgreSQL, `begins_with("f%o_o\\")` emits `f%o_o\%` as the bind parameter, but should be `f\%o\_o\\%` if `LIKE` is used, as I don't thing it is meant to handle these wildcards within it in there; the `\` at the end in this example also ends up escaping the final `%`, making it actually not be a prefix search.
4. SQL standard actually doesn't have a default escape character, PostgreSQL just defaults to `\`, maybe we want to consider being explicit with the `ESCAPE` clause?
5. PostgreSQL provides the `starts_with()` function and `^@` operator, so we could actually emit something like `WHERE starts_with(col_name, $1)` or `WHERE col_name ^@ $1)` and not worry about escape character there at all
6. Would make sense to me to have `ends_with` and `ilike` to complement the `starts_with` or `begins_with` and `like`

Issue #732, opened 2026-04-20
Express backward pagination as a driver capability


carllerche:
## Problem

DynamoDB only supports forward pagination. Its `ExclusiveStartKey` /
`LastEvaluatedKey` mechanism scans in one direction, with no built-in
"previous page" operation. SQL databases support backward pagination by
reversing the ORDER BY direction and flipping keyset comparison operators.

Today the code hardcodes this: SQL executors populate `prev_cursor`, the
DynamoDB driver sets it to `None`. The planner and `Page` API don't know
which drivers support backward pagination — applications infer it at runtime
from `prev_cursor.is_some()`.

## Proposed solution

Add a `backward_pagination: bool` field to `Capability`:

```rust
pub struct Capability {
    // ... existing fields ...

    /// Whether the driver supports backward (previous-page) pagination.
    /// SQL: true. DynamoDB: false.
    pub backward_pagination: bool,
}
```

Set `true` for all SQL drivers and `false` for `Capability::DYNAMODB`.

Planner behavior:

- **`backward_pagination` true (SQL):** the engine produces both `next_cursor`
  and `prev_cursor`. For a backward page request, it reverses ORDER BY, flips
  comparison operators in the WHERE clause, and reverses the result set
  before returning.
- **`backward_pagination` false (DynamoDB):** `ExecResponse::prev_cursor` is
  always `None`. `Page::prev()` returns `None`. Callers check
  `page.has_prev()` or `page.prev_cursor.is_some()` to decide whether to
  render a "Previous" button.

The `Page` API stays uniform — always exposing `next_cursor` and
`prev_cursor` — while drivers declare what they actually support.

## Alternatives considered

Leave the behavior hardcoded per driver as it is today. Applications can
already check `prev_cursor.is_some()` at runtime, but the planner has no
single source of truth for "does this driver support backward pagination"
and cannot branch on capability during planning.

## Scope

Small — fits in one PR, no design doc needed.

## Additional context

Follow-on from the engine-aware pagination work. The design doc at
`docs/dev/design/dynamodb-pagination.md` has been removed; the rest of that
design is already implemented.

kvasi34:
Can I work on the proposed solution?

carllerche:
> Can I work on the proposed solution?

Sure!

joeydewaal:
Implemented in #757

PR #757, opened 2026-04-25
feat: add backward_pagination driver capability

kvasi34:
## Summary

- Add a `backward_pagination` boolean field to `Capability` struct so drivers can declare whether they support backward pagination.
- Set SQL drivers `backward_pagination` value to `true` and `DynamoDB` to `false`.
- Use `backward_pagination` to determine `ExecResponse::prev_cursor` value.

Closes issue #732 

## Type of change

## Checklist

- [x] `cargo test` passes

## Notes for reviewers

The guard in `apply_sql_pagination` (`&& self.engine.capability().backward_pagination`) is defensive code for future SQL drivers that set `backward_pagination` to `false`. I believe that `DynamoDB` does not reach that part at all. Please let me know if that should exist.

There is no design doc as stated in the issue.


joeydewaal:
Could you add a driver test to assert there being/not being a previous page based on the capability? You can use `#[driver_test(requires(backward_pagination))]` and `#[driver_test(requires(not(backward_pagination)))]` in the tests.

kvasi34:
@joeydewaal I have added two tests as requested. The tests can be found in `crates/toasty-driver-integration-suite/src/tests/crud_composite_key_pagination.rs`, since that file tests similar logic. 

Both tests work in a similar fashion; navigate to the second page of a query and check whether the previous page exists.

Please review again if possible.

joeydewaal:
lgtm, thanks!

joeydewaal (approved):

PR #767, opened 2026-04-26
fix(engine): dedupe kv fan-out keys when loading BelongsTo on DDB

### LINKED ISSUE #701: DDB nested merge panics with `HashIndex: duplicate key detected` when multiple parents share a BelongsTo target
## Summary

When two sibling `BelongsTo` nested-include batches produce the same parent key (i.e. two parent rows point at the same `BelongsTo` target), the DDB nested-merge path panics with `HashIndex: duplicate key detected`. The query is legitimate — multiple sibling rows sharing a `BelongsTo` target is a normal data shape.

This was surfaced while adding DDB coverage for the fix in #699 / #691. The test was reworked to sidestep the issue, but the underlying bug remains.

## Repro

Failing driver test (add to `crates/toasty-driver-integration-suite/src/tests/relation_preload.rs`):

```rust
#[driver_test(id(ID))]
pub async fn shared_belongs_to_target_hashindex_panic(test: &mut Test) -> Result<()> {
    #[allow(dead_code)]
    struct Category {
        #[key] #[auto] id: ID,
        name: String,
        #[has_many] items: toasty::HasMany<Item>,
    }

    #[allow(dead_code)]
    struct Brand {
        #[key] #[auto] id: ID,
        name: String,
    }

    #[allow(dead_code)]
    struct Item {
        #[key] #[auto] id: ID,
        title: String,
        #[index] category_id: ID,
        #[belongs_to(key = category_id, references = id)]
        category: toasty::BelongsTo<Category>,
        #[index] brand_id: ID,
        #[belongs_to(key = brand_id, references = id)]
        brand: toasty::BelongsTo<Brand>,
    }

    let mut db = test.setup_db(models!(Category, Brand, Item)).await;

    let brand = Brand::create().name("BrandA").exec(&mut db).await?;
    let cat = Category::create()
        .name("Electronics")
        .item(Item::create().title("Phone").brand(&brand))
        .item(Item::create().title("Laptop").brand(&brand))  // <-- same brand as Phone
        .exec(&mut db)
        .await?;

    // Two items point at the same Brand — panics on DDB when the nested
    // merge builds its HashIndex over the batch-loaded brand rows.
    let loaded = Category::filter_by_id(cat.id)
        .include(Category::fields().items().brand())
        .get(&mut db)
        .await?;

    let items = loaded.items.get();
    assert_eq!(2, items.len());
    for item in items {
        assert_eq!("BrandA", item.brand.get().name);
    }

    Ok(())
}
```

Passes on Sqlite / Postgres / MySQL. Panics on DynamoDB.

> **Note:** a *single* `.include(items().brand())` (not a sibling-include case) should already reproduce this on DDB as long as two items share a brand. If the existing `nested_has_many_then_belongs_to_required` test passes on DDB today it's only because each item is given a unique brand — the bug is masked by test-data choice, not by correct behavior.

## Panic (from PR #699 CI run)

```
thread 'relation_preload::sibling_nested_includes_on_shared_prefix_non_sql::id_uuid'
panicked at crates/toasty-core/src/stmt/hash_index.rs:50:13:
HashIndex: duplicate key detected

  4: core::panicking::panic_fmt
  5: toasty_core::stmt::hash_index::HashIndex::new
             at crates/toasty-core/src/stmt/hash_index.rs:50:13
  6: toasty::engine::exec::nested_merge::<...>::{closure}
             at crates/toasty/src/engine/exec/nested_merge.rs:198:21
  7: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
             ...
  8: toasty::engine::exec::nested_merge::Exec::action_nested_merge
             at crates/toasty/src/engine/exec/nested_merge.rs:200:18
```

Full DDB job log: https://github.com/tokio-rs/toasty/actions/runs/24596348643/job/71927022881

## Analysis

`HashIndex::new` requires unique keys (see `crates/toasty-core/src/stmt/hash_index.rs:44-54`). For a `MergeQualification::HashLookup` — used for `single=true` relationships like `BelongsTo` — the index is built over the **batch-loaded child rows** (`crates/toasty/src/engine/exec/nested_merge.rs:190-200`):

```rust
stmt::HashIndex::new(values, &mi.child_projections)
```

The invariant is fine: a `BelongsTo` target has unique keys *in the database*. What's broken is the input — the child-row batch contains the same brand row twice. That happens because upstream, the parent's `brand_id` values are collected as `[A, A]` (two items, same brand) and whatever stage produces the batch result yields one row per collected key instead of one row per distinct key.

The root cause is one of:

1. **Planner emits a duplicate-tolerant batch operation for HashLookup.** `plan.rs` should dedupe the key set fed to the batch fetch (for DDB this is a `BatchGetItem`/`GetByKey` fan-out) when the downstream qualification is `HashLookup`. Keys from multiple parents naturally collide — the planner is the right place to collapse that.
2. **The DDB driver returns one row per requested key instead of one row per matched key.** If `BatchGetItem` is issued with `[{id=A}, {id=A}]` and returns `[rowA, rowA]`, the driver is stitching responses per input key rather than returning the distinct matched set. `crates/toasty-driver-dynamodb/` is worth auditing.

Hypothesis (1) is cheaper and more surgical and doesn't depend on what DDB does. It also matches what the SQL drivers already do implicitly — `WHERE id IN (A, A, B)` returns distinct rows — which is why the bug is DDB-only.

## Suggested fix

- In `plan.rs`, when lowering a nested BelongsTo into a batched child query for a `HashLookup` qualification, dedupe the key set (e.g. collect into a `BTreeSet<Value>` before handing it to the batch op).
- Keep the `debug_assert!` in `HashIndex::new` — the invariant is correct, the caller was feeding it bad data.
- Optionally, promote the `debug_assert!` to a full `assert!` so release builds surface the bug instead of silently stomping map entries.

## Workaround in PR #699

The DDB test in #699 was restructured to give each item a distinct brand + supplier (commit `4358821` on branch `claude/fix-issue-691-A3dn4`) so the HashIndex input is incidentally unique. That sidesteps this bug but doesn't fix it — any real-world query where sibling rows share a BelongsTo target will hit this on DDB.

## Related

- #691 — the sibling-include lowering bug (unrelated root cause, found while adding DDB coverage for the fix).
- #699 — the PR that fixes #691 and added the test that surfaced this.

carllerche:
## Summary

Fixes #701: on DynamoDB, a `BelongsTo` lookup where multiple parents share the same target panicked with `HashIndex: duplicate key detected` in the `NestedMerge` step.

The path that triggers the bug is `mir::QueryPk` (DDB's `Query`), not `mir::GetByKey`: `Item.brand_id -> Brand.id` is a primary-key reference but lowers to `QueryPk` with `index=None`. `Exec::action_query_pk` calls `split_filter`, which fans an `IN`-list of `[A, A]` into two `id == A` predicates and runs the same query twice, returning the same row twice — and `NestedMerge` then trips the (now strengthened) `assert!` in `HashIndex::new`. SQL drivers were unaffected because `WHERE id IN (...)` collapses duplicates implicitly.

The earlier commit on this branch added a `ValueSet` dedup in `action_get_by_key`, strengthened the `HashIndex` `debug_assert!` to `assert!`, and added a regression test. The strengthened assert correctly surfaces the bug in the test, but the dedup itself was at the wrong layer for this query shape.

This PR moves the dedup down to `split_filter_in_list` and `split_filter_any_map` so every kv-layer fan-out (`QueryPk`, `FindPkByIndex`, and the existing `GetByKey` path) collapses duplicate keys uniformly. The `action_get_by_key` dedup stays — its input variable does not flow through `split_filter`.

## Type of change

## Checklist

- [x] `cargo test` passes (full DDB suite, 294 tests, also passes)

## Notes for reviewers

The new test (`nested_has_many_then_shared_belongs_to`) reproduces the issue verbatim from #701 and now passes on every driver. The strengthened `HashIndex::new` `assert!` is intentional — kept so any future regression in dedup surfaces in release builds.


carllerche:
This should be part of #750

PR #777, opened 2026-04-26
fix: preserve non-reference constraints when lifting BelongsTo IN-subquery

joeydewaal:
## Summary
`LiftBelongsTo::visit_expr_binary_op` had a catch-all `_ => {}` arm for
constraints whose operands aren't `Expr::Reference` (e.g. an embedded
field accessed inside an `in_query`: `Project(ref(addr), [city]) = "X"`).
That left `fail = false` with `operands = []`, so the lift returned an
empty `ExprAnd` (= `true`) — silently dropping the inner WHERE clause and
making the IN subquery match every row of the target model.

Set `fail = true` in the catch-all so we fall through to the
IN-subquery form that preserves the filter verbatim.

Added a regression test (`lift_belongs_to_preserves_embedded_field_filter`)
that filters posts by an embedded field on the related user — fails on
main, passes with the fix.

## Type of change

## Checklist

- [x] `cargo test` passes

## Notes for reviewers
This was found and fixed by claude code.

### DIFF
--- a/crates/toasty-driver-integration-suite/src/tests.rs
+++ b/crates/toasty-driver-integration-suite/src/tests.rs
@@ -48,6 +48,7 @@ pub mod infra_pool_config;
 pub mod infra_reset_db;
 pub mod infra_sync_send;
 pub mod key_unsigned;
+pub mod lift_belongs_to_complex_filter;
 pub mod query_count;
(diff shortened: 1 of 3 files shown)


carllerche (approved):
Thanks! :+1:

PR #806, opened 2026-05-01
docs: filter included associations

joeydewaal:
## Summary
This design doc proposes a way to filter included associations.

[Rendered](https://github.com/joeydewaal/toasty/blob/f28631cbb5c565ffa1724297a2018816e9a3306c/docs/dev/design/include-filters.md)

## Type of change

## Checklist

- [x] `cargo test` passes

## Notes for reviewers
Any feedback is appreciated.

### DIFF
--- a/docs/dev/design/include-filters.md
+++ b/docs/dev/design/include-filters.md
@@ -10,6 +10,12 @@ writes `.include(User::fields().todos())` can write
 to preload only the unfinished todos for each user. Works the same way
 for `HasOne` / `BelongsTo` relations.
 
+Because a filtered include no longer contains the full set the relation
+field denotes, `HasMany` and `HasOne` grow a third state alongside
+`Unloaded` and `Loaded`: `Partial`. The type-level distinction keeps
+`.get()` honest — code that asks for the full set of `user.todos` cannot
+be silently handed a filtered subset.
+
 ## Motivation
 
 `include(...)` currently has no way to restrict which related records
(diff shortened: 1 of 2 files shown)


carllerche:
Thanks for writing this up — the motivation is concrete and the API reads well. Before we commit to this direction, I want to push back on it. I think `.include(...).filter(...)` quietly breaks an invariant the rest of the model API relies on.

### A relation field denotes a set, not a query result

`User::todos` means *the set of this user's todos*. The field's identity is the set it denotes — not the rows that happen to be in it after some particular load. With `.include(todos().filter(...))`, two `User` values with the same id, both with `todos` "loaded," can carry different subsets depending on which query produced them. Both look the same to the type system. Code that takes a `&User` and reads `user.todos.get()` cannot tell whether it sees all todos, the unfinished ones, the unfinished-and-pinned ones, or none-at-all-because-the-filter-excluded-them. The field's name promises one thing and contains another.

### Filtered relations as model fields

Consider `User` has many `PullRequest`s where a `PullRequest` can be closed. If "the closed pull requests of a user" is a thing the application talks about, give it a name in the schema:

```rust
#[derive(Model)]
struct User {
    #[key]
    id: Id<Self>,

    #[has_many]
    pull_requests: HasMany<PullRequest>,

    // Specify a filter somehow, though I am making up this syntax. We should probably compare with the query! design
    #[has_many(filter = PullRequest::fields().state().eq(State::Closed))]
    closed_pull_requests: HasMany<PullRequest>,
}
```

`closed_pull_requests` is a relation in its own right. It has a name, a stable meaning, and the same affordances every other relation has: `.include(User::fields().closed_pull_requests())` preloads it, `User::fields().closed_pull_requests().any(...)` works as a parent-side filter, nested includes chain through it. Nothing new in the engine — the lower phase already turns relation paths into subqueries; this gives one of those subqueries a stored predicate.

Additionally, the engine can reason about this. If both `pull_requests` and `closed_pull_requests` are eager-loaded, the engine can decide to only fetch `pull_requests` then filter in-memory.

joeydewaal:
Closing this as toasty is going with a different direction.

PR #811, opened 2026-05-02
docs: add per-call column projection design doc

MavenRain:
## Summary

This PR introduces a design doc for a `.select` method on the query builder to unlock per-call column projection.

## Type of change

- [X] Roadmap entry + design doc (no implementation in this PR)
- [ ] Other (please explain):

## Checklist

- [ ] `cargo test` passes

## Notes for reviewers
N/A

### DIFF
--- a/docs/dev/design/column-projection.md
+++ b/docs/dev/design/column-projection.md
@@ -108,9 +108,10 @@ let pairs: Vec<(u64, String)> = User::all()
 # }
 ```
 
-Tuple impls cover sizes 1 through 12, matching the standard library's
-tuple trait conventions.  For larger projections or named-field
-results, see "Open questions" below.
+Tuple impls cover sizes 1 through 10, matching the existing
+`impl_into_expr_for_tuple!` ceiling in `toasty/src/stmt/into_expr.rs`.
+For larger projections or named-field results, see "Open questions"
+below.
 
 ### Composing with other clauses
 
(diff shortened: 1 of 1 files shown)


docs/dev/design/column-projection.md
```diff
@@ -0,0 +1,475 @@
...
+    .select(Document::fields().metadata())
+    .exec(&mut db)
+    .await?;
+# Ok(())
+# }
+```
+
+### What `.select(...)` does not accept
+
+`.select(...)` rejects relation fields at compile time.  Relation
```
carllerche: Why would it reject relation fields? It seems like it should support them fine (I can't see a reason to not).

docs/dev/design/column-projection.md
```diff
@@ -0,0 +1,475 @@
...
+loading already has a dedicated surface (`.include(...)`); allowing
+relations on `.select(...)` would force the result type to bridge two
+shapes (a tuple for projections, a model+children record for relations)
+and would duplicate the include machinery.  The compile error points
+the user at `.include(...)`.
+
+`.select(())` (empty tuple) is also a compile error.  At least one
```
carllerche: I don't think it *needs* to be a compilation error.

docs/dev/design/column-projection.md
```diff
@@ -0,0 +1,475 @@
...
+
+**Selecting the same field twice.**  `.select((F, F))` is allowed; the
+column is read once and the value is duplicated into the tuple
+positions.  This matches SQL `SELECT col, col FROM ...` semantics and
+costs nothing extra to support.
+
+**Selecting a field on a model without it.**  Compile error;
+`Model::fields()` only exposes what the model declares.
+
+**Selecting through a relation traversal (`.todos().title()`).**
```
carllerche, line 329: This seems fine to support as well.

docs/dev/design/column-projection.md
```diff
@@ -0,0 +1,475 @@
...
+see operations they already handle, with a different column list.
+
+### Macro
+
+`#[derive(Model)]` generates a single new method on the query builder:
+
+- `Select<M>::select<P>(self, projection: P) -> Select<M, P::Output>`
+  where `P: Project<M>`.
+
+The `Project` trait is doc-hidden codegen support:
```
carllerche: I don't think it should be #[doc_hidden)] since it used in a public API.

docs/dev/design/column-projection.md
```diff
@@ -0,0 +1,475 @@
...
+
+```rust
+#[doc(hidden)]
+pub trait Project<M: Model> {
+    type Output: stmt::Decode;
+    fn columns(&self) -> Vec<stmt::ExprReference>;
```
carllerche: Probably `fields(&self)` since this is an app-level thing. Also, it probably shouldn't return a `Vec`. My guess is you want it to return a `stmt::Expr`.

Also, is there a reason why we can't just use `IntoExpr` and avoid defining a new `Project` trait?

docs/dev/design/column-projection.md
```diff
@@ -0,0 +1,475 @@
...
+that already underlies model record decoding; tuples and primitives
+are existing `Decode` shapes.
+
+### Engine
+
+Lowering already walks the field list and emits a column projection
+in `Returning::Project`.  The change is to consult one extra input
+from the user-facing query:
+
+- **`select_override: Option<Vec<ExprReference>>`**, when present,
```
carllerche: I am not following why `select_override` is needed at all here.

carllerche (changes requested):

carllerche (approved):

PR #814, opened 2026-05-04
feat(core): support Vec<scalar> model fields on SQLite

carllerche:
## Summary

First slice of the [document and collection fields design](https://github.com/tokio-rs/toasty/blob/docs/json-fields-design/docs/dev/design/document-fields.md): `Vec<scalar>` model fields work end-to-end on SQLite. Whole-value insert/read/update only — query operators (`.contains`, `.len`) and in-place mutations (`stmt::push`, `stmt::clear`) land in subsequent PRs.

User-facing:

```rust
#[derive(toasty::Model)]
struct Item {
    #[key] #[auto] id: i64,
    tags: Vec<String>,                 // works
}

Item::create().tags(vec!["admin".into(), "verified".into()]).exec(&mut db).await?;
```

Macro behavior is gated on traits, not type-name parsing — `Vec<T>` participates as a model field via blanket `impl<T: Field + Scalar + IntoExpr<T>> Field for Vec<T>`. `Scalar` is a sealed marker that excludes `u8` so `Vec<u8>` keeps its bytes-blob meaning.

Key pieces:

- **Storage selection** — `db::Type::{Array(Box<Type>), Json}`; `StorageTypes::scalar_list: ScalarList` picks the form per backend (PG → Array, SQLite/MySQL → Json).
- **Driver capability split** — `array_binding` (driver accepts `Value::List` as one bind parameter) and `in_array` (dialect supports `IN <bound_array>`). Independent, with `in_array` implying `array_binding`. SQLite is `(true, false)`; PG would be `(true, true)`.
- **Parameter extraction** — `extract_params` is a stateful `VisitMut` that tracks `in_list_rhs_depth` so a `Value::List` of scalars can ride as one parameter for collection-column writes while still decomposing to per-element placeholders for an `IN` rhs (when `!in_array`).
- **SQLite driver** — `Value::List` ↔ JSON text round-trip (`serde_json::Value::Array`).

PostgreSQL and MySQL drivers are not yet wired up — `Vec<scalar>` will fail at runtime on those backends until follow-up PRs implement their parameter binding. Integration tests are gated `requires(sql)` and pass on SQLite.

## Type of change

- [x] Implements a previously accepted design (link: design lives on the [`docs/json-fields-design` branch](https://github.com/tokio-rs/toasty/blob/docs/json-fields-design/docs/dev/design/document-fields.md), still in review)
- [ ] Other

## Checklist

- [x] `cargo test` passes (820/820 SQLite integration; 284/284 toasty unit)
- [x] Touches a public API → design doc covers it (link above)
- [x] Touches the `Driver` trait or driver-observable behavior → design doc covers it

## Notes for reviewers

- Two new capability flags (`array_binding`, `in_array`). Implication direction `in_array → array_binding` is **not** enforced in `Capability::validate()` yet — easy to add if desired.
- The `extract_params` refactor replaces the prior `for_each_expr_mut` + ad-hoc pre-pass with a `VisitMut` impl that carries context. The post-order traversal preserves the previous semantics; the new context counter is the only behavioral addition.
- `MYSQL` is set to `array_binding: false` to gate `Vec<scalar>` on that backend until the JSON-list bind path lands. Pre-existing MySQL functionality is unaffected.
- The two commits on the branch (`wip: scalar lists`, `refactor(engine): track IN-list context …`) collapse cleanly under squash-merge.

### DIFF
--- a/crates/toasty-core/src/driver.rs
+++ b/crates/toasty-core/src/driver.rs
@@ -20,7 +20,7 @@
 //! ```
 
 mod capability;
-pub use capability::{Capability, SchemaMutations, StorageTypes};
+pub use capability::{Capability, ScalarList, SchemaMutations, StorageTypes};
 
 mod response;
 pub use response::{ExecResponse, Rows};
--- a/crates/toasty-core/src/driver/capability.rs
+++ b/crates/toasty-core/src/driver/capability.rs
@@ -125,6 +125,27 @@ pub struct Capability {
     /// Whether the driver supports backward (previous-page) pagination.
     /// SQL: true. DynamoDB: false.
     pub backward_pagination: bool,
+
+    /// Whether the driver accepts a `Value::List` as a single bind
+    /// parameter representing a scalar array.
+    ///
+    /// PostgreSQL binds a native array. SQLite and (eventually) MySQL
+    /// bind JSON-encoded text. When `false`, scalar lists in
+    /// non-decomposable positions (e.g. a `Vec<scalar>` column value) have
+    /// nowhere to go — the schema layer is responsible for rejecting
+    /// `Vec<scalar>` fields on such backends.
+    pub array_binding: bool,
+    /// Whether `x IN <bound_array>` is a valid SQL form (e.g. PostgreSQL
+    /// `x = ANY($1)` / `x IN (SELECT unnest($1))`).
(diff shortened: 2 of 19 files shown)


carllerche:
This PR will most likely be closed.

Issue #237, opened 2025-12-04
feature: add a method to distinguish between "trusted" values and "untrusted" values.


carllerche:
Right now, when serializing statements, we assume that *every* value in the statement could be untrusted and should be escaped. In practice, this is not necessarily true. We should investigate if there is a way to track untrusted values through the system.

Exploration is needed to find prior art and to decide if doing this is actually a good idea.

joeydewaal:
`Trusted` has been renamed to `Static` (https://github.com/tokio-rs/toasty/pull/816#issuecomment-4424822975) and has been implemented in #1001.

PR #816, opened 2026-05-04
docs: add design for static SQL values

MavenRain:
## Summary

Design doc for #237.  Threads a `trust: Trust` tag through `Expr::Value` so schema-known literals stay inline as SQL while user-supplied values continue to be extracted as bind parameters.

Iteration 1 covers the type-tag plumbing, selective behavior in extract_params, the verify-pass tightening for LIMIT and OFFSET, and the simplifier-rule audit needed to keep the lattice rule ("untrusted poisons trusted") sound.  RETURNING-derived values stay Untrusted (always re-bind) until a follow-on adds the per-shape allow-list.

Prior-art survey covers Diesel, SeaORM, JOOQ, Slick, Entity Framework Core, and Ecto, with Ecto's `^pin` operator framed as the closest analogue to the explicit-marker direction Toasty declines for the user-facing API.

## Type of change

- [X] Roadmap entry + design doc (no implementation in this PR)
- [ ] Other (please explain):

## Checklist

- [ ] PR title uses [Conventional Commits](https://www.conventionalcommits.org/) format
- [ ] `cargo test` passes

## Notes for reviewers
N/A


carllerche:
Sorry for the delay. There was a bit of discussion in Discord, and I wanted to see if I could synthesize thoughts. The conclusion (and I think I agree) is mostly that this should be reframed away from "untrusted" and made `Literal` or `Static`. Even trusted values should probably be passed in as binds to better use prepared statements. "Static" values should be part of the SQL query vs. a bind. So, it isn't about trusted/untrusted but whether you want the value to be encoded as part of the SQL statement and *not* included as a bind. Trusted is implicit, though.

carllerche (approved):

PR #821, opened 2026-05-05
feat(dynamodb): add full-table scan support

### LINKED ISSUE #741: feat: DynamoDB scan support for queries with no partition key predicate
### Problem

On DynamoDB, any query that has no partition key predicate on any available index currently fails with an unsupported-feature error. This makes it impossible to fetch all records of a model (User::all()) or filter on non-key attributes without first knowing the partition key — a common and expected pattern.

 

### Proposed solution

 Introduce a new Operation::Scan variant that the planner emits when no viable index exists for a query against a NoSQL driver. The main pieces:

- plan_index_path returns Option<IndexPlan> instead of Result<IndexPlan>. The None arm in plan_data_loading_nosql dispatches to a new plan_scan() helper rather than returning an error.
- plan_scan() validates the query (no order_by — scans don't support ordering) and builds a mir::Scan node.
- The exec layer transparently follows LastEvaluatedKey to fetch all pages when no user limit is set, and threads the cursor through for explicit paginated calls.
- The DynamoDB driver implements exec_scan wrapping the AWS SDK scan() call.

### Alternatives considered

_No response_

### Scope

Medium — touches public API or one subsystem

### Additional context

_No response_

asiefker:
## Summary

- Adds `Operation::Scan` to `toasty-core` for full-table scans without a key constraint, alongside a shared `Pagination` type (extracted from `QueryPkLimit`) used by both `Scan` and `QueryPk`
- Adds `Capability::scan` and `Capability::scan_supports_sort` flags; the query planner emits `Scan` when no usable index exists and the driver declares scan support; DynamoDB sets both, SQL drivers do not
- Implements `exec_scan` in the DynamoDB driver using the AWS SDK paginator for the unbounded case (correctly crossing the 1 MB page boundary), a single `send()` for cursor pagination, and a skip+collect paginator loop for offset pagination
- Fixes a pre-existing bug where `exec_query_pk` with `limit: None` did a single `.send()` and silently truncated results at the 1 MB DynamoDB boundary; it now uses the paginator loop as well
- Adds integration tests covering filter, no-filter, limit, offset, cursor multi-page pagination, ORDER BY rejection on DDB scan, and 1 MB boundary crossing for both scan and query-pk paths

## Test plan

- [x] `cargo test` passes (SQLite)
- [x] `cargo test --features dynamodb --no-default-features -- --test-threads=1` passes against a local DynamoDB instance
- [x] New tests in `scan_no_index.rs` exercise all scan paths
- [x] New tests in `dynamodb_limit_boundary.rs` cover 1 MB boundary for `Offset`, `Cursor` (GSI), and no-limit `QueryPk`

Closes #741


carllerche (approved):
Nice!

PR #823, opened 2026-05-07
refactor(engine): move eq-operand rewrite into lower

MavenRain:
## Summary

PR #5 of the lower-and-simplify epic.  `simplify_expr_eq_operand` matches on `Reference::Model` and `BelongsTo` field references, both app-level constructs that lowering removes.  Per the design, app-level rewrites belong in lowering rather than simplify.

This PR moves the rewrite to a new `lower/expr_eq_operand.rs` module.  The new `RewriteEqOperand` struct implements `VisitMut` and fires the rewrite on every binary-op operand whose op is `eq` or `ne`.  Scope management mirrors `simplify::Simplify`: each statement-level visit pushes the source's scope onto the expression context before visiting filter and returning clauses, so `resolve_expr_reference` resolves field references against the right model.

`LoweringState::lower_stmt` and `LowerStatement::lower_sub_stmt` invoke `RewriteEqOperand` after `RewriteVia` and before the existing Simplify pre-lower pass.  After this PR, `simplify` no longer rewrites `Reference::Model` or `BelongsTo` operands; the "Move to lower" list still has three items: `lift_in_subquery`, `rewrite_expr_in_list_when_model`, and `try_variant_tautology_or`.

## Type of change

- [X] Implements a previously accepted design
- [ ] Other (please explain):

## Checklist

- [X] `cargo test` passes

## Notes for reviewers

N/A


crates/toasty/src/engine/lower/expr_eq_operand.rs
```diff
@@ -0,0 +1,154 @@
...
+        // operands rewritten before this node's operands are inspected.
+        stmt::visit_mut::visit_expr_binary_op_mut(self, i);
+
+        if i.op.is_eq() || i.op.is_ne() {
+            self.rewrite_operand(&mut i.lhs);
+            self.rewrite_operand(&mut i.rhs);
+        }
+    }
+
+    fn visit_stmt_delete_mut(&mut self, stmt: &mut stmt::Delete) {
```
carllerche: A bunch of this looks like it is duplicating the lower visitor. Is that intentional? Could it be combined *with* the `lower` visitor logic? If not, perhaps we should look for ways to avoid that duplication? (I'm referencing stuff like this in lower https://github.com/tokio-rs/toasty/blob/main/crates/toasty/src/engine/lower.rs#L550).

carllerche (approved):
thanks

PR #753, opened 2026-04-24
docs: add standalone CLI design doc

joeydewaal:
## Summary

Adds a design doc for making `toasty-cli` usable as a standalone installed binary without requiring users to write a custom CLI binary per project.
I think this is a big improvement for developer experience as you can `cargo install toasty-cli` and just work with the migrations.

[Rendered](https://github.com/joeydewaal/toasty/blob/be2a9b6a8417a0f9a142402d95d51b89aafe793a/docs/dev/design/standalone-cli.md)

## Type of change

- [ ] Small / obvious change — bug fix, docs, internal cleanup, or test
- [ ] Implements a previously accepted design
      (link to the merged design doc under `docs/dev/design/`):
- [x] Roadmap entry + design doc (no implementation in this PR)
- [ ] Other (please explain):

## Checklist

- [x] PR title uses [Conventional Commits](https://www.conventionalcommits.org/) format
- [x] `cargo fmt` and `cargo clippy` are clean
- [x] `cargo test` passes
- [x] Touches a public API → a design doc under `docs/dev/design/` describes the behavior
- [x] Touches the `Driver` trait or driver-observable behavior → the design doc covers it

## Notes for reviewers

A big part of the logic/idea behind having the cli set an env var and build the user's project has been taken from the sqlx-cli from the SQLx crate. Their `offline` feature has been using this mechanism for quite some time.

cc: @pikaju

PR #824, opened 2026-05-07
feat: make cli standalone

### LINKED ISSUE #762: Standalone CLI: collect schema by compiling a synthesized dumper crate instead of via proc-macro side effects
## Problem

#753 proposes a standalone `toasty-cli` that drives migrations without per-project boilerplate. The collection mechanism in that design is a compile-time side effect: when `TOASTY_SCHEMA_OUT` is set, `#[derive(Model)]` writes each model's schema to a TOML file during `cargo check`, and the CLI assembles those files into the full schema.

That mechanism does not work for the full Toasty type graph. A proc macro sees only source tokens, not resolved types. It cannot tell that `field: SomeAlias` resolves to an `Embed` defined in another crate, that `BelongsTo<Foo>` refers to model `path::to::Foo` after re-exports, or that a generic parameter resolves to a particular embedded type. Type aliases, re-exports, and `cfg`-gated paths each break the static-resolution assumption. The same limitation is the reason `typeshare` documents loss of cross-crate type information, and the reason `specta` chose runtime trait info instead.

Toasty already has the information needed to produce a fully resolved schema. `#[derive(Model)]` registers each type with `inventory` (`crates/toasty-macros/src/model/expand/model.rs:101`), and `Register::register` walks embedded and relation fields recursively, calling back into rustc-resolved types. A standalone CLI that can run any code linked against the user's models gets the resolved schema for free.

## Proposed solution

`toasty migrate` synthesizes an ephemeral "dumper" crate under the user's `target/` directory, builds it, runs it, and reads the serialized schema from its stdout.

Layout:

```
<user>/
  Cargo.toml           # untouched
  src/main.rs          # may contain models inline
  target/
    toasty-dump/
      Cargo.toml       # generated
      src/dumper.rs    # generated
```

The generated `target/toasty-dump/Cargo.toml`:

```toml
[package]
name = "toasty_dump"
version = "0.0.0"
edition = "2021"

[lib]
path = "../../src/main.rs"   # or src/lib.rs if the user has a lib
crate-type = ["rlib"]

[[bin]]
name = "toasty-dumper"
path = "src/dumper.rs"

[dependencies]
# mirrored from `cargo metadata` for the user's package

[workspace]   # marks this crate as its own workspace root
```

The generated `src/dumper.rs`:

```rust
fn main() {
    let mut set = toasty::schema::ModelSet::new();
    for item in toasty::codegen_support::inventory::iter::<toasty::DiscoverItem>() {
        item.add_to(&mut set);
    }
    let schema = toasty::schema::from_macro(set).unwrap();
    serde_json::to_writer(std::io::stdout(), &schema).unwrap();
}
```

The CLI:

1. Runs `cargo metadata` to find the target package and its dependencies.
2. Writes the synthetic `Cargo.toml` and `dumper.rs` into `target/toasty-dump/`, mirroring the user's `[dependencies]`, `[features]`, `edition`, and `rust-version`.
3. Runs `cargo build --manifest-path target/toasty-dump/Cargo.toml`.
4. Executes the resulting binary and parses its stdout.

This produces the fully resolved `app::Schema` regardless of how complex the relation or embedded-type graph is, because rustc has already done the resolution.

### Handling bin-only crates

When the user's package is `[[bin]]`-only (no `[lib]`), the synthetic `[lib] path` points at `src/main.rs` directly. Rust treats any `.rs` file as a library root; `mod foo;` resolves relative to `src/`, the same way it does in the bin build, so module discovery still works. A `fn main` in the file is just a regular function in lib context.

### Notes

- The empty `[workspace]` table marks the synthetic crate as its own workspace root so that the user's workspace globs cannot absorb it.
- `models!(crate::*)` filters `DiscoverItem`s by `CARGO_PKG_NAME`. In the dumper, that name is `toasty_dump`, which also matches what `#[derive(Model)]` recorded when expanding under the synthetic lib. The dumper sidesteps the question by iterating every `DiscoverItem` directly — only one user-source root is linked, so there is nothing else in the inventory.
- First build is a full compile; subsequent runs are incremental within `target/toasty-dump/target/`. Cached registry sources are reused.

## Alternatives considered

**Proc-macro side effect (the design in #753).** Discussed above. Cannot resolve types that the proc macro does not see literally in the token stream.

**Inject a temporary `examples/` target into the user's crate.** Same idea but writes into the source tree. Mutates `Cargo.lock`, requires cleanup on failure, and breaks the "no source-tree writes" property.

**User opt-in dumper helper.** Toasty exposes `toasty::dump_schema_if_requested()` and the user calls it early in `main`; the CLI runs the user's existing binary with `TOASTY_DUMP_SCHEMA=path` set. Avoids synthesizing a crate but requires the user to add a call site, which defeats the goal of zero per-project boilerplate.

**`cargo rustc -- --crate-type=lib` on the user's bin.** Produces an `.rmeta`/`.rlib` for the bin compiled as a lib without a synthetic `Cargo.toml`. The dumper would then need its own Cargo.toml and hand-assembled `--extern` flags for the entire transitive closure. Brittle compared to letting Cargo resolve deps for the synthetic crate.

**Static section with `linkme`.** A `linkme::distributed_slice` of `&'static SchemaFragment` lives in the binary as inert data and could be parsed out with `object`/`goblin` without running user code. Requires every part of `app::Model` to become `const`-constructible (`&'static str`, `&'static [...]`, no `OnceLock` for `ModelId`, no recursive trait dispatch during construction). Large refactor for a small win over running a dumper binary.

**Static source analysis (typeshare-style).** Reimplements rustc's name resolution. Loses information across crates, fails on type aliases and `cfg`. Same wall the proc-macro approach hits.

## Scope

Large — new subsystem, cross-cutting with the macro crate, and a new entry point in the CLI.

## Additional context

- #753 — the static collection design this issue responds to.
- `crates/toasty/src/schema/register.rs` — current `DiscoverItem` / `inventory` collection.
- `crates/toasty-macros/src/model/expand/model.rs:101` — where `#[derive(Model)]` emits `inventory::submit!`.
- Prior art for the synthesized-crate pattern:
  - [`cargo-fuzz`](https://github.com/rust-fuzz/cargo-fuzz) — generates a sibling Cargo project that depends on the user's crate. Requires a `[lib]`, so does not handle bin-only.
  - [`rust-script`](https://crates.io/crates/cargo-script) — synthesizes a fresh Cargo package in a managed cache directory outside the source tree.
  - [`cargo-mutants`](https://github.com/sourcefrog/cargo-mutants) — uses `cargo metadata` and writes derivative artifacts under `target/`.
- Related but different goals:
  - [`specta`](https://github.com/specta-rs/specta) — runtime trait info for type export.
  - [`typeshare`](https://crates.io/crates/typeshare-cli) — static analysis; documents loss of cross-crate info.

joeydewaal:
## Summary

With this pr it's possible to `cargo install toasty-cli` and use it as a standalone cli-tool. Before this wasn't possible and users were required to write their own tool on top of the `toasty-cli` crate.

Closes #762

## Type of change

## Checklist

- [x] `cargo test` passes

## Notes for reviewers
The `snapshot` command was removed. I'm not quite sure what it's use is and it was easier to just remove it all together.
Also this moves `toasty-cli` more towards a cli rather than a library.
This changes the `Driver` trait and removes the `generate_migration` method on it. This allows the cli to generate migrations without requiring an open database connection.

See #762 and #753 for more context about the implementation.


carllerche:
Superseded by #831

PR #825, opened 2026-05-07
fix: reverse OrderByExpr direction correctly for backward pagination

### LINKED ISSUE #804: Backward pagination returns wrong previous page order
### Summary

I was working on issue #732 when I noticed the following.

Retrieving the previous page using `Page::prev()` when in `ASC` order returns a page with elements in `DESC` order.

#### Expected behavior

Check the reproducer code. My understanding is that when

```
page1 = 0 1 2 3 4 5 6 7 8 9
page2 = 10 11 12 13 14 15 16 17 18 19
```

then `page2.prev()` should be the same as `page1`.

```
prev = 0 1 2 3 4 5 6 7 8 9 (should be)
```

Is my understanding correct? Shouldn't that be the output when calling `Page::prev()` and order is `ASC`?

#### Actual behavior

However, the actual values in `page2.prev()` are

```
prev = 19 18 17 16 15 14 13 12 11 (actual)
```

#### Why does this happen?

Backward pagination works by reversing the ORDER BY, fetching the page, then reversing the result set.  
Because the direction never flips for ascending queries, the whole chain goes wrong:
                                                                                                        
1. Query is `ORDER BY seq ASC`           
2. `prev()` calls `before(cursor)` which sets `reverse = true`
3. `exec()` calls `order_by.reverse()` but this doesn't change `ASC` to `DESC`. `ASC` stays `ASC` in `crates/toasty-core/src/stmt/order_by_expr.rs` `reverse` method.

### Reproducer

```rust
#[key(partition = kind, local = seq)]
struct Event {
    kind: String,
    seq: i64,
}

let driver = toasty_driver_sqlite::Sqlite::in_memory();
let mut db = toasty::Db::builder()
    .models(toasty::models!(Event))
    .build(driver)
    .await?;
db.push_schema().await?;

// Seed 20 events under the same partition key.
for i in 0..20_i64 {
    Event::create().kind("info").seq(i).exec(&mut db).await?;
}

// Page 1: seq 0–9 (ascending)
let page1: toasty::stmt::Page<_> = Event::filter_by_kind("info")
    .order_by(Event::fields().seq().asc())
    .paginate(10)
    .exec(&mut db)
    .await?;

println!("page1 ({} items):", page1.len());
for e in page1.iter() {
    println!("  seq={}", e.seq);
}

// Page 2: seq 10–19 (ascending)
let page2: toasty::stmt::Page<_> = page1.next(&mut db).await?.unwrap();

println!("\npage2 ({} items):", page2.len());
for e in page2.iter() {
    println!("  seq={}", e.seq);
}

// Go back to page 1 via prev()
let prev: toasty::stmt::Page<_> = page2.prev(&mut db).await?.unwrap();

let prev_seqs: Vec<i64> = prev.iter().map(|e| e.seq).collect();
let expected: Vec<i64> = (0..10).collect();

println!("\nprev ({} items):", prev.len());
for e in prev.iter() {
    println!("  seq={}", e.seq);
}

println!("\nExpected: {expected:?}");
println!("Got:      {prev_seqs:?}");
```

### Database driver

SQLite, PostgreSQL, MySQL

### Toasty version

commit 5b1e09e

### Additional context

Running the reproducer codes gives the following console output.

```
page1 (10 items):
  seq=0
  seq=1
  seq=2
  seq=3
  seq=4
  seq=5
  seq=6
  seq=7
  seq=8
  seq=9

page2 (10 items):
  seq=10
  seq=11
  seq=12
  seq=13
  seq=14
  seq=15
  seq=16
  seq=17
  seq=18
  seq=19

prev (9 items):
  seq=19
  seq=18
  seq=17
  seq=16
  seq=15
  seq=14
  seq=13
  seq=12
  seq=11

Expected: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
Got:      [19, 18, 17, 16, 15, 14, 13, 12, 11]
```

kvasi34:
## Summary

- Fix `OrderByExpr::reverse` not changing `Asc` to `Desc`.
- Modified `crud_composite_key_pagination::paginate_composite_key_prev` test to check the number and order of items in the previous page (this covers previous pages in `Asc` order).
- Added `crud_composite_key_pagination::paginate_composite_key_prev_desc` test to cover previous pages in `Desc` order.

Closes #804 

## Type of change

## Checklist

- [x] `cargo test` passes

### DIFF
--- a/crates/toasty-core/src/stmt/order_by_expr.rs
+++ b/crates/toasty-core/src/stmt/order_by_expr.rs
@@ -26,12 +26,11 @@ pub struct OrderByExpr {
 }
 
 impl OrderByExpr {
-    /// Flips the sort direction. `Desc` becomes default (ascending); default
-    /// and `Asc` become `Asc` (explicit ascending, since the default changed).
+    /// Flips the sort direction. `Asc` becomes `Desc` and `Desc` becomes default (ascending).
     pub fn reverse(&mut self) {
         self.order = match self.order {
-            Some(Direction::Desc) => None,
-            _ => Some(Direction::Asc),
+            Some(Direction::Asc) => Some(Direction::Desc),
+            _ => None,
         }
     }
 }
--- a/crates/toasty-core/src/stmt/order_by_expr.rs
+++ b/crates/toasty-core/src/stmt/order_by_expr.rs
@@ -26,11 +26,12 @@ pub struct OrderByExpr {
 }
 
 impl OrderByExpr {
-    /// Flips the sort direction. `Asc` becomes `Desc` and `Desc` becomes default (ascending).
+    /// Flips the sort direction. `Desc` becomes `Asc`; default (ascending)
+    /// and `Asc` become `Desc`.
     pub fn reverse(&mut self) {
         self.order = match self.order {
-            Some(Direction::Asc) => Some(Direction::Desc),
-            _ => None,
+            Some(Direction::Desc) => Some(Direction::Asc),
+            _ => Some(Direction::Desc),
         }
     }
 }
--- a/crates/toasty-driver-integration-suite/src/tests/crud_composite_key_pagination.rs
+++ b/crates/toasty-driver-integration-suite/src/tests/crud_composite_key_pagination.rs
@@ -367,6 +367,51 @@ pub async fn paginate_composite_key_prev(test: &mut Test) -> Result<()> {
     let prev: Option<Page<_>> = page2.prev(&mut db).await?;
     assert!(prev.is_some());
 
+    // Check the number and order of the items in the previous page
+    let prev_page = prev.unwrap();
+    assert_eq!(prev_page.len(), 10);
+    for i in 0..10 {
+        assert_eq!(prev_page[i].seq, i as i64);
+    }
+
+    Ok(())
+}
+
+#[driver_test(requires(backward_pagination))]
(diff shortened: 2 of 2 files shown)


crates/toasty-core/src/stmt/order_by_expr.rs
```diff
@@ -26,12 +26,11 @@ pub struct OrderByExpr {
...
 impl OrderByExpr {
-    /// Flips the sort direction. `Desc` becomes default (ascending); default
-    /// and `Asc` become `Asc` (explicit ascending, since the default changed).
+    /// Flips the sort direction. `Asc` becomes `Desc` and `Desc` becomes default (ascending).
     pub fn reverse(&mut self) {
         self.order = match self.order {
-            Some(Direction::Desc) => None,
-            _ => Some(Direction::Asc),
+            Some(Direction::Asc) => Some(Direction::Desc),
+            _ => None,
```
joeydewaal: Since the default/`None` is ascending shouldn't `None` be reversed to `Direction::Desc`?

kvasi34: You are right, @joeydewaal . That was a silly mistake on my part. I fixed the `reverse` method. It does the following now.

- `Desc` -> `Asc`
- `Asc` -> `Desc`
- `None` -> `Desc`

Please let me know if everything is okay now.

joeydewaal (approved):

PR #831, opened 2026-05-08
docs: design standalone CLI via linktime ctor + cdylib

carllerche:
## Summary

[Rendered](https://github.com/tokio-rs/toasty/blob/worktree-bright-moseying-penguin/docs/dev/design/standalone-cli.md)

Proposes an alternative to the synthesized-dumper-crate scheme in #824. A constructor in `toasty` (gated on `debug_assertions`) dumps the schema and exits when an env var is set; `toasty-cli` builds the user's existing target — a `[[bin]]` directly, or the `[lib]` as a `cdylib` via `cargo rustc --crate-type cdylib` — and triggers the ctor by running it. The cdylib path uses a hidden re-exec subcommand to host the dlopen so the ctor's `exit(0)` does not kill the parent.

This drops the synthesized companion crate, removes the manifest-mirroring requirement, and picks up bin-only crates that the synthesized-crate approach does not support. Workspace package selection is via `-p`, mirroring `cargo`.

The design has one open question that is blocking acceptance: do constructors fire reliably when built via `cargo rustc --crate-type cdylib` on every supported platform? This PR ships a smoke test (`tests/tests/cdylib_ctor_smoke.rs`) that builds a fixture as a cdylib, dlopens it, and asserts a `#[linktime::ctor]` wrote its marker. CI's `test-all-os` job runs the test on `ubuntu-latest`, `macos-latest`, and `windows-latest`.

The PR is opened as a draft to gather the cross-platform CI signal before proceeding.

Refs #762, #824.

## Type of change

## Checklist

- [x] \`cargo fmt\` and \`cargo clippy\` are clean
- [x] \`cargo test\` passes (locally on macOS; Linux/Windows pending CI)
- [x] Touches a public API → a design doc under \`docs/dev/design/\` describes the behavior

## Notes for reviewers

The smoke test is the load-bearing part of this PR. If it passes on all three platforms, the design's open question resolves and implementation can proceed. If it fails on a platform, the design needs a per-platform fallback for the cdylib path.

The design doc otherwise stands on its own and supersedes the older proc-macro side-effect proposal in #753.


joeydewaal:
@carllerche whats the status on this/can I take over?

PR #837, opened 2026-05-09
feat: add a basic axum example

darricheng:
## Summary

Add a basic example for using Toasty with Axum.

I was looking for one as a base to use and play with but couldn't find one, so I added this example that would hopefully be of use to others as well.

## Type of change

- [ ] Other (please explain):

## Checklist

- [x] `cargo test` passes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ members = [
     "examples/hello-toasty",
     "examples/todo-with-cli",
     "examples/user-has-one-profile",
+    "examples/simple-axum-server",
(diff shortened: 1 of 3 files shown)


carllerche:
Thanks for the contribution. I think the conclusion is this would make more sense in the Axum repo as they already have examples with ORMs. I would follow up with them first to see if they would take the example. cc @joeydewaal

darricheng:
@carllerche yup that makes sense, thanks for taking a look at the PR!

Issue #830, opened 2026-05-08
feat: item collection / single-table design support for DynamoDB


asiefker:
### Problem

DynamoDB users have no ergonomic path for modeling hierarchical data today. A typical multi-tenant app with Tenant → User → Todo → Subtask must either:

- Create one DDB table per model type, giving up DDB's single-partition co-location and paying multiple query round trips for hierarchy loads.
- Hand-roll composite sort keys and a discriminator column in raw attributes, forfeiting Toasty's relation API.

Single-table design is the DDB-native answer — one table, partition key = root id, sort key = "<Prefix>#<id>#<Prefix>#<id>…" — but it requires orchestration across the macro, schema, engine, and driver layers that Toasty does not provide.

Affected: Toasty users writing DDB-backed applications with hierarchical data. Drivers in-tree (DDB implementation work; SQL drivers add a rejection path).

### Proposed solution

Add `#[item_collection]` as a storage-layout directive that maps a model hierarchy onto one DDB table.

```rust

  #[derive(Model)]
  #[item_collection]
  #[key(partition = id, local = sk)]
  struct Tenant {
      #[auto]
      id: Uuid,
      #[version]
      version: u64,

      name: String,
      #[has_many]
      users: HasMany<User>,
  }

  #[derive(Model)]
  #[item_collection(Tenant)]
  #[key(partition = tenant_id, local = sk)]
  struct User {
      #[auto]
      id: Uuid,
      tenant_id: Id<Tenant>,
      #[version]
      version: u64,

      name: String,
      #[belongs_to]
      tenant: BelongsTo<Tenant>,
      #[has_many]
      todos: HasMany<Todo>,
  }

  #[derive(Model)]
  #[item_collection(User)]
  #[key(partition = tenant_id, local = sk)]
  struct Todo {
      #[auto]
      id: Uuid,
      tenant_id: Id<Tenant>,
      #[version]
      version: u64,

      title: String,
      #[belongs_to]
      user: BelongsTo<User>,
  }
```

### User-facing surface:

- Scoped creation is the only way to create children: tenant.users().create()…. Unscoped User::create() is not generated for item-collection children.
- Id<T> stays a bare Uuid; the composite sort key is engine-internal and never exposed.
- .include() collapses to a single DDB Query when all included relations live in the same collection: User::filter_by_id(u).include(todos()).include(todos().subtasks()) issues one call.
[...]

PR #869, opened 2026-05-11
docs: design for item collections on DynamoDB

asiefker:
## Summary

  [Rendered](https://github.com/tokio-rs/toasty/pull/869/changes#diff-53704a17c28d64867c60693ab3baf1373d9cdd7c08fac655da7434621bb0b38c)

  Adds a guide-level design doc for `#[item_collection]`, a DynamoDB-only
  storage directive that maps a model hierarchy onto a single DDB table. Also
  adds a roadmap entry under "Relationships" linking to the doc and issue.

  Key points of the design:

  - `#[item_collection]` on the root, `#[item_collection(Parent)]` on children.
    Intermediate ancestor ids are not declared as fields; the sort key encodes
    the ancestry chain, and Toasty manages it internally.
  - Scoped creation is mandatory for children. Top-level `Todo::create()` and
    `Todo::get_by_*` are not generated for item-collection children.
  - `.include()` chains that stay within a collection collapse into one DDB
    `Query`. All-equality chain navigation collapses into one `GetItem`.
  - Cascade deletes are explicitly out of scope on DynamoDB: the 100-item
    `TransactWriteItems` cap prevents all-or-nothing semantics for arbitrary
    collection sizes, and silent divergence from SQL's cascade atomicity is
    worse than no cascade.
  - SQL drivers reject item-collection schemas at `push_schema` with a clear
    error.

  Refs #830.

  ## Type of change

  ## Checklist

  ## Notes for reviewers

  Two open questions in the doc need explicit feedback before acceptance:

  1. **Breaking change for DDB users with composite-key schemas.** Multi-field
     `#[key]` attributes get rejected by the DDB driver at `push_schema`; users
     must migrate to struct-level `#[key(partition = …, local = …)]`. SQL
     drivers are unaffected. Toasty is pre-1.0; the break is on the table but
     worth a deliberate yes/no.
  2. **Deep-hierarchy over-fetch cost envelope.** `HasMany` queries on deep
     trees over-fetch descendants and discard them client-side. RCU cost should
     be quantified before v1 decides whether a GSI-backed opt-out ships as part
     of v1 or waits.

### DIFF
--- a/docs/dev/design/item-collections.md
+++ b/docs/dev/design/item-collections.md
@@ -1,0 +1,546 @@
+# Item Collections
+
+## Summary
+
+An `#[item_collection]` model attribute declares that a model participates in a
+single DynamoDB table shared with a related model. Toasty synthesizes a composite
+sort key from the ancestry chain, collapses `.include()` and chain-navigation
+calls into single DynamoDB requests, and rejects item-collection schemas at
+connection time on SQL drivers. Item-collection is a DynamoDB-only storage
+directive; the Rust API is otherwise unchanged.
+
+## Motivation
+
+DynamoDB single-table design co-locates hierarchical data under one partition
+key to get cheap subtree reads and fewer round trips on hierarchy loads.
+Toasty users with a `Tenant → User → Todo → Subtask` shape today have two poor
+options: one table per model with an access pattern that pays N queries for
+hierarchy loads, or hand-rolled sort keys and discriminators that skip
+Toasty's relation API entirely.
+
+The single-table idiom requires coordination across macros, schema, engine, and
+driver: composite sort-key synthesis on write, prefix-based filter translation
+on read, per-row model dispatch when a query returns multiple row types. The
+plumbing does not exist today. This design adds it.
+
+Affected:
+
+- **Toasty users** writing DynamoDB-backed applications with nested relations.
+- **DynamoDB driver maintainers** (implementation target).
+- **SQL driver maintainers** (add one rejection path at schema setup; no other
+  changes).
+
+## User-facing API
+
+### Declaring an item collection
+
+Mark every model that participates in the shared table with
+`#[item_collection]`. The root uses the bare form; children name their direct
+parent.
+
+```rust
+#[derive(Debug, toasty::Model)]
+#[item_collection]
+#[key(partition = id, local = sk)]
+struct Tenant {
+    #[auto]
+    id: Uuid,
+
+    name: String,
+
+    #[version]
+    version: u64,
+
+    #[has_many]
+    users: HasMany<User>,
+}
+
+#[derive(Debug, toasty::Model)]
+#[item_collection(Tenant)]
+#[key(partition = tenant_id, local = sk)]
+struct User {
+    #[auto]
+    id: Uuid,
+
+    tenant_id: Id<Tenant>,
+
+    name: String,
+
+    #[version]
+    version: u64,
+
+    #[belongs_to]
+    tenant: BelongsTo<Tenant>,
+
+    #[has_many]
+    todos: HasMany<Todo>,
+}
+
+#[derive(Debug, toasty::Model)]
+#[item_collection(User)]
+#[key(partition = tenant_id, local = sk)]
+struct Todo {
+    #[auto]
+    id: Uuid,
+
+    tenant_id: Id<Tenant>,
+
+    title: String,
+
+    #[version]
+    version: u64,
+
+    #[belongs_to]
+    user: BelongsTo<User>,
+}
+```
+
+Three things to notice:
+
+- **Every child declares the root's id as its partition key field**
+  (`tenant_id: Id<Tenant>`). All rows in an item collection share the same
+  partition key. Intermediate ancestor ids — a Subtask's `user_id` or
+  `todo_id` — are *not* declared. Ancestry lives inside the sort key, which
+  Toasty manages; models never expose or construct it.
+
+- **`Id<T>` is a bare `Uuid`.** Same meaning as on standalone models. There is
+  no opaque compound identifier type for item-collection members.
+
+- **`#[belongs_to]` arguments are optional.** For item-collection children the
+  macro uses the ancestry chain from `#[item_collection(Parent)]` to derive
+  `key` and `references` from the partition-key field and the parent's primary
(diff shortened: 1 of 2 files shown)


docs/dev/design/item-collections.md
```diff
@@ -0,0 +1,546 @@
+# Item Collections
+
+## Summary
+
+An `#[item_collection]` model attribute declares that a model participates in a
+single DynamoDB table shared with a related model. Toasty synthesizes a composite
+sort key from the ancestry chain, collapses `.include()` and chain-navigation
+calls into single DynamoDB requests, and rejects item-collection schemas at
```
carllerche, line 8: There's no need to limit this to DynamoDB. It may not be very useful on SQL, but it can be implemented. That said, this seems like a method to represent a tree structure in a single table. You could use a similar strategy to store a homogeneous tree as well as(e.g., users). I think it is worth considering this as well. Maybe it shouldn't be called `#[item_collection]` in Toasty and something reflecting the tree nature of the pattern?

docs/dev/design/item-collections.md
```diff
@@ -0,0 +1,546 @@
...
+### Declaring an item collection
+
+Mark every model that participates in the shared table with
+`#[item_collection]`. The root uses the bare form; children name their direct
+parent.
+
+```rust
+#[derive(Debug, toasty::Model)]
+#[item_collection]
+#[key(partition = id, local = sk)]
```
carllerche, line 44: Here, `local` refers to a "hidden" field that contains the hierarchical path tracked by the item collection. Maybe we should have a way to reference that field, like `$item_path` or something. Or maybe, the field name can be named in `#[item_collection]` something like `#[item_collection(path = foo)]` then `#[key(partition = id, local = foo)]`

docs/dev/design/item-collections.md
```diff
@@ -0,0 +1,546 @@
...
+    #[version]
+    version: u64,
+
+    #[has_many]
+    users: HasMany<User>,
+}
+
+#[derive(Debug, toasty::Model)]
+#[item_collection(Tenant)]
+#[key(partition = tenant_id, local = sk)]
```
carllerche, line 60: I don't think you need to specify the key here necessarily as it should be inferred when this is linked to the `Tenant` collection.

docs/dev/design/item-collections.md
```diff
@@ -0,0 +1,546 @@
...
+    id: Uuid,
+
+    tenant_id: Id<Tenant>,
+
+    name: String,
+
+    #[version]
+    version: u64,
+
+    #[belongs_to]
```
carllerche, line 72: This `#[belongs_to]` is also doing a lot of the heavy lifting with regard to the item collection. Maybe we can do `#[item_parent]` or something instead of `#[belongs_to]` (and drop `#[item_collection(Tenant)]`).

carllerche (commented):
Looks pretty good. The main thing I would like to figure out is how to frame this as a more general feature than ddb-specific. I had some more specific API thoughts inline as well.

docs/dev/design/item-collections.md
asiefker: I agree that item_collection is DDB specific. The key idea is using a single table to model a tree. Some alternatives: 

* `#[single_table]`
* `#[colocate]`
* ??? 

I'll add others if I come up with anything.

docs/dev/design/item-collections.md
asiefker: As I was writing this up, I was split between: 
1. Reusing existing concepts like `belongs_to` and `has_many`
2. How redundant the modeling was. An item collection implies the relationship. 

I'll iterate on some examples.

docs/dev/design/item-collections.md
asiefker: I think that's a good simplification. The `User` struct would continue to declare them, though correct?

docs/dev/design/item-collections.md
asiefker: How would `$item_path` be used?  I'm not opposed to declaring `foo` in `item_collection`, but what do we get by referring to it in both `item_collection` and `key`? 

PR #939 kept the sort key as a standard field. In the database, it's value is `<Type>#<parent1 id>#<parent2 id>#....<parentN id>`. On load from the db, the sk is parsed, and the local id is set on the struct.

docs/dev/design/item-collections.md
asiefker: Consolidating this feedback into a single example, I think we get something like: 

```rust
#[derive(toasty::Model)]
#[key(account, sk)]
struct Tenant {
    account: String,
    #[auto] sk: String,
    name: String,
    #[has_many] users: Deferred<Vec<User>>,
}

#[derive(toasty::Model)]
struct User {
    account: String,
    sk: String,
    name: String,
    #[item_parent] tenant: Deferred<Tenant>
    #[has_many] todos: Deferred<Vec<Todo>>,
}

#[derive(toasty::Model)]
struct Todo {
    account: String,
    sk: String,
    title: String,
    #[item_parent] user: Deferred<User>
}
```
* #[key] is used to declared the key attributes of the root object. All children need to declare the same 2 fields, it will be a build time error if that doesn't happen. The children will implicitly treat the keys as #[auto]. 
*  item_parent replaces item_collection and belongs_to. item_parent is a union of those 2 features. 

If Toasty clients need to control id generation then `#[local_id]` can be used: 
```rust
#[derive(toasty::Model)]
struct User {
    account: String,
    sk: String,
    #[local_id] user_num: i64,
    name: String,
    #[item_parent] tenant: Deferred<Tenant>
}
```
In this case sk is the full sort key value. Toasty will split off the tail of the sort key and populate user_num on gets. On puts, Toasty will build the User's sort key from the the parent's key + user_num.

asiefker: Once we get to an approved API design, I'll update this design doc.

asiefker: While implementing the above change I realized that inferring keys across toasty::Models will not work. During macro expansion Toasty only has visibility into the current Model, so inferring keys from the parent will not work. 

Solution: All toasty::Models in the collection will need to declare #[key]. Toasty will check at schema build time that all the keys are defined exactly the same, if not there will be an error. 

There is some precedent for this: belongs_to has to do something similar when validating the parent side of the relationship.

Issue #188, opened 2025-11-16
Reconsider equality with `null` in SQL databases


pikaju:
SQL has difficult to follow rules about how `NULL` contaminates expressions (see https://en.wikibooks.org/wiki/Structured_Query_Language/NULLs_and_the_Three_Valued_Logic). Because of this a comparison operator used with `NULL` (e.g. `table.column = NULL`) is often not what you want. You have to use `table.column IS NOT NULL` instead. [`IS DISTINCT FROM`](https://wiki.postgresql.org/wiki/Is_distinct_from) is also very useful. In Rust, `NULL` does not exist and is instead represented with `Option<T>` in Toasty. However, a user has to be aware of the aforementioned quirks.

Toasty has [some code to mitigate this](https://github.com/tokio-rs/toasty/blob/05437e1a9af1fc4eef5e3e3e3abec9c74eb8db71/crates/toasty/src/engine/lower.rs#L560). But as far as I can tell there is still an issue with nullable columns, which way or may not be null and thus cannot be lowered to `IS NOT NULL`. An example of this would be joining two tables with a nullable column, which would not return rows containing `NULL` values.

carllerche:
Concrete reproducer for the `S-needs-repro` label. `.eq(None)` on a nullable field matches zero rows — including the rows that are actually `None`:

```rust
// Document { metadata: Option<Metadata> }
Document::filter(Document::fields().metadata().eq(None)).exec(&db).await?   // 0 rows
Document::filter(Document::fields().metadata().is_none()).exec(&db).await?  // 1 row (correct)
```

`None` lowers to a literal `NULL`, and `fold_expr_binary_op` (`crates/toasty/src/engine/fold/expr_binary_op.rs`) folds `expr <op> null → null`, so the predicate is the constant `NULL` and matches nothing. `.ne(None)` is symmetric. The nullable-comparison mitigation this issue links doesn't apply — this is a literal `None` on the right-hand side, not a nullable-column self-comparison. It holds for any `Option<T>` field (scalar, embed, enum); it surfaced while adding `Option<Embed>` ([#1021]).

[...]

PR #876, opened 2026-05-12
docs: design for null-safe equality on Option<T> fields

MavenRain:
## Summary

  Design doc for #188.  Proposes Option-aware `.eq` / `.ne` on nullable fields
  and `.sql_eq` / `.sql_ne` as the explicit SQL-3-value escape hatch, plus a
  simplifier and evaluator audit so SQL backends and the DynamoDB driver agree
  on what `NULL` comparisons return.

  Two new `Expr` variants (`IsNotDistinctFrom`, `IsDistinctFrom`) serialize to
  dialect-native operators: `IS NOT DISTINCT FROM` on PostgreSQL, `IS` on
  SQLite, `<=>` on MySQL.  No new `Driver` capability or schema migration
  required.

  ## Design highlights

  - Two-tier API.  `.eq` / `.ne` route on the static Rust type (`Option<T>` vs
  `T`) at the call site; `.sql_eq` / `.sql_ne` always emit raw `=` / `<>`.
  - Engine stays a strict superset of SQL.  `BinaryOp::Eq` keeps three-valued
  semantics; the new variants carry the null-safe behavior.
  - No engine-wide nullability typing.  Dispatch happens at the call site where
  Rust still has the `Option` information.
  - Behavior change on two surfaces: `nullable_field.eq(other_nullable_field)`
  now treats `(NULL, NULL)` as a match, and `nullable_field.eq(None)` now
  consistently finds `NULL` rows.  `.is_none()` / `.is_some()` are unchanged.

  ## Open questions

  The doc calls out three decisions for review:

  1.  Naming of the strict-SQL methods (`.sql_eq` vs `.eq_strict` vs `.eq_sql`
  vs `.eq_3vl`).
  2.  Whether to gate the `.eq` behavior flip behind a one-release deprecation
  that warns on nullable-pair `.eq` sites.
  3.  Whether to surface `.is_distinct_from` directly on `Path` or keep it
  expression-builder-only.

  Refs #188.

### DIFF
--- a/docs/dev/design/null-equality.md
+++ b/docs/dev/design/null-equality.md
@@ -1,0 +1,492 @@
+# Equality with `NULL`
+
+## Summary
+
+`.eq` and `.ne` on a nullable field become null-safe: comparing two
+`Option<T>` columns counts a row where both sides are `NULL` as a match,
+and comparing a nullable field to `None` is the same as
+`.is_none()`.  This matches the `Option::eq` behavior that Rust callers
+already assume.  A new `.sql_eq` / `.sql_ne` pair stays available for
+the strict SQL three-valued-logic semantics when the caller wants them.
+
+The simplifier and the in-memory evaluator are tightened in the same
+change so SQL backends and the DynamoDB driver agree on what `NULL`
+comparisons return.  Two `Expr` variants — `IsNotDistinctFrom` and
+`IsDistinctFrom` — appear in the IR and serialize to
+`IS NOT DISTINCT FROM` / `IS DISTINCT FROM` on PostgreSQL, `IS` / `IS
+NOT` on SQLite, and `<=>` / `NOT (... <=> ...)` on MySQL.  No new
+`Driver` capability is required.
+
+## Motivation
+
+Issue [#188] flags that `.eq` on a nullable column today follows SQL's
+three-valued logic, which surprises Rust callers and silently drops
+rows.  The two cases the issue raises are common in practice:
+
+- A join keyed on a nullable column — for example, `author_email =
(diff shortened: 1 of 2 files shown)


docs/dev/design/null-equality.md
```diff
@@ -0,0 +1,492 @@
...
+    .await?;
+# Ok(())
+# }
+```
+
+`.ne` is symmetric: it returns the rows where the two sides differ,
+counting `(NULL, value)` and `(value, NULL)` as differing and counting
+two `NULL`s as equal.
+
+### Strict SQL equality with `.sql_eq` and `.sql_ne`
```
carllerche: I would remove this for now. My guess is we will have a different way to inject SQL semantics.

carllerche: In general, I think it is important to not say "SQL semantics" since it might not be backed by SQL. Maybe "the semantics of the target database".

docs/dev/design/null-equality.md
```diff
@@ -0,0 +1,492 @@
...
+### Error mapping
+
+No new error variants are introduced.  The in-memory evaluator no
+longer errors on `NULL AND x`, `NULL OR x`, or comparing `NULL` with
+an ordered operator; it returns `Value::Null` (meaning unknown)
+instead, which the surrounding `WHERE` machinery handles uniformly.
+
+## Edge cases
+
+### `IN` with a `NULL` element
```
carllerche, line 237: This seems bad.

MavenRain:
As discussed with Carl, we are going to hold off on this one until we can work through the types of exprs at app-level during lowering and safely translate to SQL

PR #895, opened 2026-05-15
fix: update sqlite URL in docs to one that works

cshaa:
## Summary

Trying to connect to a database with the URL `sqlite://mydb.sqlite`, for example, doesn't work – it doesn't fail, but also doesn't create any database file (because `Url::parse("sqlite://mydb.sqlite")?.path()` is empty). One needs to connect to `sqlite:./mydb.sqlite`. Update the docs to reflect that.

## Type of change

- [ ] Other (please explain):

## Checklist

- [ ] `cargo test` passes

### DIFF
--- a/docs/guide/src/database-setup.md
+++ b/docs/guide/src/database-setup.md
@@ -57,7 +57,7 @@ Examples:
 .connect("sqlite::memory:")
 
 // SQLite file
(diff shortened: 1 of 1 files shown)


carllerche:
Thanks

PR #900, opened 2026-05-15
feat: add #[document] Vec<Embed> collection fields

carllerche:
## Summary

First increment of [`docs/dev/design/document-fields.md`](../blob/main/docs/dev/design/document-fields.md):
`Vec<T: Embed>` model fields under an explicit `#[document]` attribute,
stored as one document column (`jsonb` on PostgreSQL, JSON1 text on SQLite)
with whole-value CRUD. Each element encodes as a JSON object keyed by the
embed's field names; `None` fields are omitted.

Gated on a new `document_collections` capability — MySQL and DynamoDB are
out of scope this increment; element predicates, `partial!` containment,
per-element mutation, `#[document]` on bare embeds / `HashSet` / scalars,
and `#[document(text)]` remain follow-on work.

## Type of change

- [x] Implements a previously accepted design: [`docs/dev/design/document-fields.md`](../blob/main/docs/dev/design/document-fields.md)

## Checklist

- [x] PR title uses Conventional Commits format
- [x] `cargo test` passes (SQLite 888, PostgreSQL 916, MySQL 893, DynamoDB negative + capability validation)
- [x] Touches the `Driver` trait or driver-observable behavior → covered by the design

## Notes for reviewers

A few design choices worth attention:

- **`#[document]` is explicit, not inferred** — the macro resolves the field
  through a new `Document` trait (separate from `Field` to dodge the `Vec<T>`
  coherence clash, since no type is both `Scalar` and `Embed`).
- **`stmt::Type::Document(TypeDocument)`** carries field names through schema
  build; the builder rewrites macro-emitted `Type::List(Model(id))` to
  `Type::List(Document(..))` in `resolve_document_types` once every embed is
  registered.
- **`stmt::Value::Object` is the write-side bind representation.** The engine
  converts `Value::Record` → `Value::Object` in `extract_params::refine` using
  the column's `TypeDocument` field names; the driver encodes the named value
  structurally without needing the schema. On read the driver already has the
  expected type, so `value_from_json` decodes straight to `Value::Record` —
  `Value::Object` is unused on the read path. Asymmetric but each side does
  the minimal correct thing.
- **`Value::Record is_a Type::Document`** and the analogous `is_subtype_of`
  case make the positional record satisfy a document type — same shape with
  names dropped.


nick9822:
@carllerche is this staged for any upcoming release? Looking forward to `#Document` change. I have created `jsonb` storage type and with new value variant `Value::Jsonb(serde_json::Value)` for my use case. If this if going to merge, I can get rid of it.

carllerche:
@nick9822 `#[document]` landed on origin/main via another PR (this path was abandoned). Check that out and let me know if it covers your needs.

carllerche:
#1028 handles this instead.

PR #923, opened 2026-05-18
fix(macros): compose `#[serialize]` and `#[deferred]` on one field

### LINKED ISSUE #920: #[serialize(json)] + #[deferred] on the same field is silently broken (codegen branches don't compose)
### Summary

Marking a #[serialize(json)] field as #[deferred] is accepted by the macro validator but produces uncompilable code. The attribute-compatibility check in model/schema/field.rs rejects #[deferred] combined with relations, #[version], or #[key], but not with #[serialize]. However, the codegen branches for the two attributes are mutually exclusive and the serialize branch always wins (it appears earlier in the match), so #[deferred] becomes a no-op for the setter/loader codegen while still wrapping the field type in Deferred<T>.

### Reproducer

```rust
  pub struct Repository {
      #[key]
      #[auto]
      pub id: uuid::Uuid,

      pub name: String,

      #[serialize(json)]
      #[deferred]
      pub schema: toasty::Deferred<MyJsonType>,
  }

  where MyJsonType implements Serialize + Deserialize.

  Observed errors (abridged)

  error[E0277]: the trait bound `Deferred<MyJsonType>: serde::Deserialize<'de>` is not satisfied
     --> models.rs
      |
      | #[derive(Debug, toasty::Model)]
      |                 ^^^^^^^^^^^^^
      = note: this error originates in the derive macro `toasty::Model`

  error[E0308]: mismatched types
     = note: expected enum `toasty::schema::app::Model`
              found struct `Statement<MyJsonType>`
      (the generated `fn schema(&self) -> Statement<...>` collides with model-registration code)

  error[E0599]: no method named `clone` found for struct `Deferred<T>`
  error[E0609]: no field `columns` on type `Deferred<MyJsonType>`

  (Plus a handful of consequent errors at call sites that assume the field is still a plain T.)
```

### Database driver

SQLite

### Toasty version

0.6.1

### Additional context

_No response_

carllerche:
## Summary

Marking a field with both `#[serialize(json)]` and `#[deferred]` was
accepted by the validator but produced uncompilable code: every
codegen site matched on `serialize` first and silently dropped the
deferred branch, so setters demanded `Deferred<T>` instead of `T`,
the loader tried to JSON-decode into the wrapper, and there was no
lazy-load path through the `Load` trait for the user's serde type.

Walk through each site (setter, loader, post-update reload,
lazy-load accessor, field path) so the two attributes compose:

- Setters accept `<Deferred<T> as Defer>::Inner` and JSON-encode as before.
- The whole-model loader peels the `Null` / `Record([loaded])`
  envelope, JSON-decodes the inner column, and wraps in `Deferred::from`.
- Post-update reload does the same wrap.
- The lazy-load accessor returns `Statement<LoadJson<T>>` — a new hidden
  `Load` adapter (Output = T) that pulls a JSON String and deserializes
  via `serde_json`. Maps DB `Null` to JSON `"null"` so `Option<T>`
  decodes as `None`.
- `.include()` gets a `String`-typed path matching the JSON storage type.

Shared helpers (`serialize_setter_input_ty`, `expand_serialize_decode`,
`wrap_in_deferred`) live in a new `expand/codec.rs` so each cross-cutting
policy is defined once. This also dedupes the existing serialize-only
JSON decode block across `model.rs` and `update.rs`.

Closes #920

## Type of change

## Checklist

- [x] `cargo test` passes

## Notes for reviewers

Regression coverage in `crates/toasty-driver-integration-suite/src/tests/deferred_field.rs`
shares a new `deferred_serialize_document` scenario (`Repository` model with a
`#[serialize(json)] #[deferred] payload: Deferred<Payload>` field) across five
focused tests: create-returns-loaded, default-leaves-unloaded, exec-lazy-loads,
include-eager-loads, update-refreshes. `requires(sql)` is kept on the new tests
since the combination hasn't been exercised against the DynamoDB driver yet.


carllerche:
This PR isn't quite working as I want it to. See https://github.com/tokio-rs/toasty/issues/920#issuecomment-4478886781

PR #930, opened 2026-05-19
feat: allow #[version] on tuple-newtype embeds of u64

### LINKED ISSUE #839: macro rejects #[version] on newtype embed of u64
## Problem

The `#[version]` field check in `crates/toasty-macros/src/model/schema/field.rs` requires the field type to be syntactically `u64`:

```
error: #[version] can only be applied to a u64 field
```

That rejects newtype wrappers like `Version(u64)`, even though the user might want a type-distinct version counter for the same reason they want a type-distinct ID. PR #836 made `#[auto]` work for newtype embeds (a tuple-newtype around an `Auto` type is itself `Auto` via the `NewtypeOf` blanket). The same shape should work for `#[version]`.

Reproducer:

```rust
#[derive(toasty::Embed)]
struct Version(u64);

#[derive(toasty::Model)]
struct Doc {
    #[key]
    id: u64,
    #[version]
    version: Version,
    title: String,
}
```

Today this fails at the macro layer with the message above.

## Proposed solution

Mirror the `NewtypeOf` mechanism from #836:

- Drop the syntactic `is_u64` check in the field validator.
- Add a `Versionable` (or similarly named) trait in `codegen_support`, with a concrete `impl Versionable for u64` and a blanket `impl<T: NewtypeOf<Inner: Versionable>> Versionable for T`.
- Have the `Model` derive emit a trait obligation on the `#[version]` field type: `_check::<#field_ty>()` where `_check<T: Versionable>()`. The Rust type checker enforces that the field type resolves to `u64` (directly or through any depth of newtype embed), with a clean diagnostic via `#[diagnostic::on_unimplemented]`.
- The schema-builder propagation that PR #836 set up for `#[auto]` would extend to `#[version]` the same way (single-field embed flattens the outer's `versionable` flag down to the inner column).

## Alternatives considered

- **Keep the gap.** Users lose type-distinctness for version counters but the API stays smaller. Workable if no one asks for it.
- **Explicit `#[version]` attribute on the embed.** Mirrors the pre-#836 explicit `#[auto]` design; rejected for the same reason — implicit promotion via `NewtypeOf` is cleaner once the blanket exists.

## Scope

## Additional context

Discussed in #836; left out of that PR because the dead schema-builder propagation would have made the unreachable.

Pritam-deb:
## Summary

Closes #839 

A `#[version]` field can now be a tuple-newtype embed wrapping `u64` rather than a bare `u64` — no extra annotation required:

```rust
#[derive(Debug, Clone, Copy, PartialEq, toasty::Embed)]
struct Version(u64);

#[derive(toasty::Model)]
struct Item {
    #[key]
    #[auto]
    id: uuid::Uuid,

    name: String,

    #[version]
    version: Version,
}

```

The macro previously hard-rejected any non-`u64` syntactic type on `#[version]`. The guard is replaced by a `Versionable` trait in `codegen_support` with a concrete impl for `u64` and a blanket that propagates through `NewtypeOf` embeds at any depth, mirroring the existing Auto blanket from #836.

`#[derive(Model)]` emits a `_check::<FieldTy: Versionable>()` obligation for every `#[version]` field so invalid types produce a clean compiler error via `#[diagnostic::on_unimplemented]`.

The generated OCC logic is updated to call `Versionable::as_u64` / `from_u64` instead of treating the field as a bare `u64`. The insert lowering reuses the existing `auto_target` wrap-depth logic to initialize the version to the correct record shape for embedded types. The DynamoDB driver is extended to unwrap single-field record expressions in filter conditions, which is how a newtype field reference appears after schema mapping lowering.

## Type of change

- [ ] Other (please explain):

## Checklist

- [x] `cargo test` passes

## Notes for reviewers

Four DynamoDB integration tests are added mirroring the existing `field_version.rs` scenarios. They are marked `requires(not(sql))` since the SQL drivers do not yet implement the version OCC path.
The `Expr::Record` arm added to `ddb_expression` only handles single-field records. Multi-field embedded structs in filter position hit an explicit `todo!` — they are not yet supported and I found no path that produces them.
`NewtypeOf` gains a `from_inner` constructor (alongside the existing `into_inner`) so generated code can reconstruct a wrapper value from a raw `u64` without knowing the concrete type.

### DIFF
--- a/crates/toasty-driver-dynamodb/src/lib.rs
+++ b/crates/toasty-driver-dynamodb/src/lib.rs
@@ -418,6 +418,18 @@ fn ddb_expression(
             let inner = ddb_expression(cx, attrs, primary, &expr.expr);
             format!("size({inner})")
         }
+        // A single-field record is a newtype embed (e.g. `Version(u64)`).
+        // DynamoDB stores the inner value as a flat attribute, so unwrap one
+        // level and recurse — `Record([Column(c)])` is equivalent to `Column(c)`.
+        stmt::Expr::Record(expr_record) => {
+            let [inner] = expr_record.fields.as_slice() else {
+                todo!(
+                    "multi-column embedded struct in DynamoDB filter: {:#?}",
+                    expr_record
+                );
+            };
+            ddb_expression(cx, attrs, primary, inner)
+        }
         _ => todo!("FILTER = {:#?}", expr),
     }
 }
--- a/crates/toasty-driver-dynamodb/src/lib.rs
+++ b/crates/toasty-driver-dynamodb/src/lib.rs
@@ -418,18 +418,6 @@ fn ddb_expression(
             let inner = ddb_expression(cx, attrs, primary, &expr.expr);
             format!("size({inner})")
         }
-        // A single-field record is a newtype embed (e.g. `Version(u64)`).
-        // DynamoDB stores the inner value as a flat attribute, so unwrap one
-        // level and recurse — `Record([Column(c)])` is equivalent to `Column(c)`.
-        stmt::Expr::Record(expr_record) => {
-            let [inner] = expr_record.fields.as_slice() else {
-                todo!(
-                    "multi-column embedded struct in DynamoDB filter: {:#?}",
-                    expr_record
-                );
-            };
-            ddb_expression(cx, attrs, primary, inner)
-        }
         _ => todo!("FILTER = {:#?}", expr),
     }
 }
--- a/crates/toasty-macros/src/model/expand/update.rs
+++ b/crates/toasty-macros/src/model/expand/update.rs
@@ -191,8 +191,8 @@ impl Expand<'_> {
 
         quote! {
             {
-                let current: u64 = <#field_ty as #toasty::Versionable>::as_u64(s.target.#field_ident);
-                let next = <#field_ty as #toasty::Versionable>::from_u64(current + 1);
+                let current: u64 = <#field_ty as #toasty::Version>::as_u64(s.target.#field_ident);
+                let next = <#field_ty as #toasty::Version>::from_u64(current + 1);
                 s.assignments.set(
                     #toasty::stmt::Projection::from_index(#index_tokenized),
                     <#field_ty as #toasty::IntoExpr<#field_ty>>::into_expr(next),
--- a/crates/toasty-macros/src/model/expand/update.rs
+++ b/crates/toasty-macros/src/model/expand/update.rs
@@ -197,6 +197,7 @@ impl Expand<'_> {
                     #toasty::stmt::Projection::from_index(#index_tokenized),
                     <#field_ty as #toasty::IntoExpr<#field_ty>>::into_expr(next),
                 );
+                let current_val = <#field_ty as #toasty::Version>::from_u64(current);
                 s.condition = Some(
                     #toasty::core::stmt::Expr::eq(
                         #toasty::core::stmt::Expr::Reference(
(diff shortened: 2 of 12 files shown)


crates/toasty-driver-dynamodb/src/lib.rs
```diff
@@ -421,6 +418,18 @@ fn ddb_expression(
             let inner = ddb_expression(cx, attrs, primary, &expr.expr);
             format!("size({inner})")
         }
+        // A single-field record is a newtype embed (e.g. `Version(u64)`).
+        // DynamoDB stores the inner value as a flat attribute, so unwrap one
+        // level and recurse — `Record([Column(c)])` is equivalent to `Column(c)`.
+        stmt::Expr::Record(expr_record) => {
```
carllerche: This shouldn't actually happen. In theory, the engine should remove the record from new types before passing them on. I'm not sure why it isn't.

crates/toasty/src/codegen_support/version.rs
```diff
@@ -0,0 +1,61 @@
...
+/// A field type that can be used as an OCC version counter.
+///
+/// Only `u64` and tuple-newtype embeds that wrap a `Versionable` type (directly
+/// or transitively) satisfy this bound.
+#[diagnostic::on_unimplemented(
+    message = "`{Self}` cannot be used as a `#[version]` field",
+    label = "invalid type for `#[version]`",
+    note = "only `u64` and tuple-newtype embeds of `u64` are supported"
+)]
+pub trait Versionable: Copy {
```
carllerche: nit: we don't use `*able` trait names. `trait Version` is probably fine.

crates/toasty-driver-dynamodb/src/lib.rs
carllerche: It looks like the issue is in `expand_version_update_stmts`, it needs to be updated to accomodate new-type versions. My guess is the `Version` trait needs a method to increment the version as well.

edit: actually, maybe not?

crates/toasty-macros/src/model/expand/update.rs
carllerche: Ok, I think *this* is the problem. It is setting the `eq` condition hard coded to u64 when it should be `#field_ty` into_expr()

crates/toasty/src/codegen_support/version.rs
Pritam-deb: this is fixed now, renamed it across.

crates/toasty-driver-dynamodb/src/lib.rs
Pritam-deb: I agree, `Expr::Record` in the driver is a workaround of what should be handled upstream in the engine. I actually added it to get the tests passing but you are right that the record wrapping around a newtype field reference should be stripped before reaching the driver.

Happy to remove it and track down where in the lowering the unwrap should happen, or leave that as a follow-up if you'd prefer to keep the scope of this PR narrow.

crates/toasty-macros/src/model/expand/update.rs
Pritam-deb: Makes sense, but that was the earlier state. The current code uses `Version::from_u64` and `IntoExpr<FieldTy>` to build the condition expression in the correct shape for the field type, so it should be producing the right value for both bare `u64` and newtype wrappers now. Let me know if you see anything still off there.

carllerche: No, it is still there. You did update part of the expr to use `Version::from_u64` but there is still a hard coded. I think that is causing the record to show up

Pritam-deb: i see it now, the RHS of OCC condition was hardcoded as `Value::U64(current)`, the assignment was fixed to use `IntoExpr<FieldTy>`, but the condition comparison wasn't.
I guess this shape mismatch is what caused the `Expr::Record` to reach the DynamoDB driver — the engine can't simplify the comparison because the two sides have different shapes?

crates/toasty-driver-dynamodb/src/lib.rs
Pritam-deb: removed this block of written code.

carllerche (approved):
LGTM :+1: thanks.

PR #939, opened 2026-05-21
feat: item collection

asiefker:
## Summary
First revision for DynamoDB Single Table Support

This change adds the item_collection attribute that enables multiple toasty::Models to be stored in a single DynamoDB table under a parent's partition key. 

This first draft is only for DynamoDB. 
`examples/item-collection/src/main.rs` shows how to create and query items from a single table. 

All user facing APIs should be supported. 
Delete is currently not atomic. Deletes use the current behavior of belongs_to/has_many. Children will be deleted one by one and not with a transaction. There will be follow up work to address delete specifically. 

This change does not address the feedback from the design PR. I wanted to get this out for review, then I'll address the design comments and update this PR as necessary. 

Draft implementation of PR #869 

## Type of change

- [ ] Other (please explain):

## Checklist

- [x] `cargo test` passes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ members = [
     # Examples
     "examples/composite-key",
     "examples/hello-toasty",
+    "examples/item-collection",
     "examples/todo-with-cli",
     "examples/user-has-one-profile",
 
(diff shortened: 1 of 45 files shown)


carllerche:
I'm going to try to review this. At first, I saw 8k+ diff and I almost died, but it looks like most of that is in expand.rs in example. Probably worth removing that. 😆

asiefker:
I’ll update the review request.

asiefker:
Closing in favor of #1038

PR #945, opened 2026-05-22
fix(dynamodb): encode bool values as N("1"/"0") to match key attribute type

### LINKED ISSUE #856: bug(dynamodb): `bool` value in filter rejected as `Invalid attribute value type`
### Summary

A query with a bool literal in the filter predicate fails on DynamoDB with `ValidationException: Invalid attribute value type`. The bool is being bound as a DynamoDB `BOOL` attribute, but the column appears to be encoded differently (likely `N`).

### Reproducer

`crates/toasty-driver-integration-suite/src/tests/crud_query_macro.rs:248` (`query_macro_filter_bool_literal`):

```rust
struct Item {
    #[key]
    #[auto]
    id: ID,
    name: String,
    #[index]
    active: bool,
}

toasty::create!(Item::[
    { name: "on", active: true },
    { name: "off", active: false },
])
.exec(&mut db)
.await?;

let items = toasty::query!(Item filter .active == true).exec(&mut db).await?;
```

### Additional context

```
service error: unhandled error (ValidationException):
  code: "ValidationException"
  message: "Invalid attribute value type"
```

The field is `#[index]` so the path goes through `FindPkByIndex`, not `Scan`. So this bug pre-dates the scan landing (#821) and is independent of it — surfaced while reviewing `requires(sql)` tests.

Pritam-deb:
## Summary

Fixes a type mismatch that caused DynamoDB to reject filter predicates on indexed `bool` fields with `ValidationException: Invalid attribute value type`.

DynamoDB only allows `S`, `N`, or `B` for key attribute types; `BOOL` is rejected. The driver was encoding all `stmt::Value::Bool` as `AV::Bool`, but key attributes were declared as `N` — a mismatch DynamoDB rejects at the API level.

Rather than special-casing Bool in the driver, the fix belongs in the schema build phase where the driver's constraints are already applied:

- Add `Capability::bool_key_type` (false for DynamoDB, true for all SQL backends).
- `build_model_constraints` sets `FieldPrimitive::storage_ty = db::Type::Integer(1)` for any Bool field that is a primary-key or GSI key attribute when `!bool_key_type`.
- `db::Type::bridge_type` maps `(Integer(1), Bool) → I8`, so the column's stmt-level type becomes `I8`. The engine's existing cast paths handle `Bool ↔ I8` transparently.
- The driver uses native `AV::Bool` for non-key Bool fields and sees `Value::I8` for key Bool fields — no special-casing needed anywhere in the driver.
- Removed the pre-existing `Bool => ScalarAttributeType::N` workaround from `type.rs`; key Bool fields now arrive as `I8` at the driver.

Closes #856

## Type of change

## Checklist

- [x] `cargo test` passes

## Notes for reviewers

I have tried using AI for this PR, and all the code generated was reviewed carefully before commit.


carllerche:
Doesn't dynamodb support bool? I would think we should use ddb bools. I am not sure exactly what is going on here that caused the problem in the first place. cc @asiefker

Pritam-deb:
@carllerche Yes you are right, dynamoDB does support native BOOL for non-key attributes; the restriction is only that key attributes (primary key / GSI) must be S, N, or B.

From what I understood the root cause was a mismatch: `type.rs` already declares `Bool → N` when defining key attributes, but `value.rs` encoded all bools as `AV::Bool`. The uniform N fix keeps them in sync at the cost of losing native BOOL for non-key fields.


The more precise fix would be to use `AV::Bool` for non-key bools and `AV::N` for key bools, but `Value::to_ddb()` has no column context to tell those apart. Happy to rework it to thread a `is_key` flag through if you'd prefer — `from_ddb()` would also need to accept both `AV::Bool` and `AV::N` for `Type::Bool` to handle any existing data stored with the old encoding.

carllerche:
I'm hoping @asiefker can weigh in as he is the resident ddb expert :)

asiefker:
I think it would be best if Toasty can support DynamoDb's boolean type for non-key values. For boolean key attributes, I think the conversion to AV:N should also be maintained. 

I wonder if this translation makes more sense to do during the planning or lowering phase instead of inside the DDB driver? Those phases have full access to schema metadata, so should be able to emit the correct `stmt::Type` - `stmt::Type::Bool` or `stmt::Type::I8` depending on context. Then the DDB driver just does the transform without additional context.  

I will spend time investigating this today and report back. We can then coordinate who will make the actual change.

asiefker:
I think it makes the most sense to update [build_model_constraints](https://github.com/tokio-rs/toasty/blob/2772512a75206c9d1799bcb2787905137b3b3b76/crates/toasty-core/src/schema/builder.rs#L184). It has access to the database specific Capability as well as the app::Model, so it can make decisions based on the database. 

Model expansion time will not work, as Toasty doesn't know which database will be used during compile time. The logic could be put in the driver, but Toasty already does a type mappings during `build`, and I think it makes sense to keep that there. 

@carllerche and @Pritam-deb -  What do you think of this approach?

Pritam-deb:
@asiefker can `build_model_constraints`  see which fields are key/index attributes at the point it runs (before columns exist), since that information lives on `app::Model::Root.indices`?
on the other hand there i found [create_column](https://github.com/tokio-rs/toasty/blob/2772512a75206c9d1799bcb2787905137b3b3b76/crates/toasty-core/src/schema/builder/table.rs#L993) which runs per-column with the full table context.
We need to know whether a field is a key/index attribute, which is in `app::Model::Root.indices`.
Both can see it, but `create_column` is closer to the point of column creation.

Pritam-deb:
Overall I do think that @asiefker your architectural direction is on point!
The schema build phase has the right context and is already where similar decisions live — bridge_type handles Uuid→Bytes and Text→String for the same reason: the engine should work with a normalized type rather than the driver needing special cases everywhere.

From this I understand the mechanism would look like this: 

> For a bool column that is a key/index attribute on a driver that can't use BOOL keys, produce db::Type::Integer(1) as the storage type and let bridge_type return stmt::Type::I8. The driver then sees stmt::Value::I8(1) / stmt::Value::I8(0) and encodes them as AV::N("1") / AV::N("0") without any special bool handling. Non-key bool columns keep db::Type::Boolean → bridge_type → stmt::Type::Bool → AV::Bool. No driver changes needed at all.


I do have few clarifications that i require from your side; refering to your investigation, i have asked one question already.

asiefker:
> see which fields are key/index attributes at the point 

For base tables, [Field](https://github.com/tokio-rs/toasty/blob/667868eb47d099efa3ed2a3b8b9f4e2a914703a8/crates/toasty-core/src/schema/app/field.rs#L43) declares `primary_key: bool`, which  is true if the field represents a primary key. 

For GSIs, [IndexFields](https://github.com/tokio-rs/toasty/blob/667868eb47d099efa3ed2a3b8b9f4e2a914703a8/crates/toasty-core/src/schema/app/index.rs#L92) index_scope will tell you if the field is the partition key in an index. IndexField::field can be compared to Field::id to match them up. 

## build_model_constraints vs create_column 
I looked at both `build_model_constraints` and `build_column` this morning. We can make either location work: 
* build_model_constraints would update `FieldPrimitive::storage_ty` based on the database, field type, and primary key status. This is nice because it centralizes storage overrides in 1 field. This field can either be updated at compile time via `#[column(type=XXX)` or at schema build time. I would not update `db::Type::from_app` in that case. It may make sense to refactor `from_app` to take a `FieldPrimitive` instead of the components
* [create_column](https://github.com/tokio-rs/toasty/blob/667868eb47d099efa3ed2a3b8b9f4e2a914703a8/crates/toasty-core/src/schema/builder/table.rs#L993) is the other place. The indexes can be accessed via something like: `self.build.app.models.get(&self.build.mapping.id).unwrap().as_root_unwrap().indices`.  

I prefer the change to `build_model_constraint` via `FieldPrimitive`because type overrides stay in 1 place.

Pritam-deb:
@carllerche i think this PR is ready to be reviewed

crates/toasty-core/src/schema/builder.rs
```diff
@@ -175,13 +176,47 @@ impl Default for Builder {
 impl BuildSchema<'_> {
     fn build_model_constraints(&self, model: &mut app::Model) -> Result<()> {
         let model_name = model.name().to_string();
+
+        // Collect field indices of Bool fields used as key/index attributes.
+        // These need db::Type::Integer(1) as their storage type on backends
+        // that don't support BOOL as a key attribute type (e.g. DynamoDB).
+        let bool_key_field_indices: HashSet<usize> = if let app::Model::Root(root) = &*model {
```
asiefker: I generally prefer immutable variables, but I think in this case the code will be easier to follow with: 

```rust 
let mut bool_key_field_indices = HashSet::new();
if let ..... {

// populate bool_key_field_indices as needed. 
}
// delete both else blocks.
```

I also find `matches!(&f.ty, app::FieldTy::Primitive(stmt::Type::Bool))` a bit easier to read. 

Instead of id.index, I would build the HashSet out of FieldId. I prefer the descriptive  FieldId to usize. Not a huge issue here as we do just use the set locally.

crates/toasty-driver-dynamodb/src/lib.rs
```diff
@@ -433,6 +433,16 @@ fn ddb_expression(
...
             format!("size({inner})")
         }
+        stmt::Expr::Cast(expr_cast) => {
+            // Bool key/index fields bridge through I8 (db::Type::Integer(1)).
+            // The lowering wraps the I8 column ref in Cast(col_ref, Bool) when
+            // the field appears as a bare predicate after `field = true`
+            // simplification. Emit an explicit equality against the stored N("1").
+            let col_alias = ddb_expression(cx, attrs, primary, &expr_cast.expr);
+            let true_val =
+                attrs.ddb_value(aws_sdk_dynamodb::types::AttributeValue::N("1".to_string()));
```
asiefker, line 445: There's not a `field = false` use case here? 

More generally, there's no way for a Cast to be for something other than a boolean? I can see how that might be the case today, but always emitting a `true` value seems to broad. Can we narrow this so there's an error if Cast is for something Toasty doesn't support?

asiefker (changes requested):
I think this looks pretty good. 

For the first comment, those are my personal preferences, but the suggestions are not blockers. I do think the second comment/question is the more important one.

crates/toasty-core/src/schema/builder.rs
Pritam-deb: noted the suggestion and applied. Thanks.

crates/toasty-driver-dynamodb/src/lib.rs
Pritam-deb: applied the changes in the lastest commit.

asiefker (approved):

PR #959, opened 2026-05-25
fix(engine): handle ExprOr in eval verify_expr

### LINKED ISSUE #857: bug(engine): `ExprOr` not implemented in eval for nested AND/OR with comparison ops
### Summary

A filter of the form `eq AND (gt OR lt)` panics during evaluation with `not yet implemented: expr=ExprOr {..}` at `engine/eval.rs:98`. The flat `OR` case (`eq OR eq` on the same field) works because the simplifier rewrites it to `IN`; the comparison-OR-inside-AND case does not get rewritten and falls through to the missing eval branch.

### Reproducer

`crates/toasty-driver-integration-suite/src/tests/crud_query_macro.rs:139` (`query_macro_filter_parens`):

```rust
toasty::create!(User::[
    { name: "Alice", age: 30 },
    { name: "Bob", age: 20 },
    { name: "Carl", age: 40 },
])
.exec(&mut db)
.await?;

// .name == "Alice" AND (.age > 25 OR .age < 15)
let users = toasty::query!(User filter .name == "Alice" and (.age > 25 or .age < 15))
    .exec(&mut db)
    .await?;
```

### Additional context

```
panicked at crates/toasty/src/engine/eval.rs:98:14:
not yet implemented: expr=ExprOr { operands: [ExprBinaryOp { .. }, ExprBinaryOp { .. }] }
```

Reproduces on DynamoDB scan path. Tests that flatten to `IN` (`name == X OR name == Y`) work; tests that mix `OR` of comparison ops inside an `AND` don't.

Probably surfaces on every driver that takes the same eval path — the comparison-OR variant just hadn't been exercised before scan opened up these tests for DynamoDB.

Pritam-deb:
## Summary

Fixes #857.

`verify_expr` in `engine/eval.rs` had an arm for `And` but not `Or`. Filters of the form `field == X AND (field > Y OR field < Z)` would recurse into the `And`, encounter the inner `ExprOr`, and hit the `todo!()` catch-all — panicking at runtime across all drivers.

`OR` of equalities was unaffected because the simplifier rewrites them to `IN` before eval. `OR` of comparison operators has no such rewrite and exposed the gap.

Fix: one line — `Or` mirrors `And` exactly, since both hold a flat list of operand expressions.

Also removes the `requires(sql)` gate from `query_macro_filter_parens`, which was added as a workaround for this bug.

## Type of change

## Checklist

- [x] `cargo test` passes

## Notes for reviewers

`verify_expr` gates whether an expression is safe to evaluate in-process as a post-fetch filter. The fix belongs here rather than in the simplifier because `OR` of comparison ops has no canonical rewrite — there's nothing to normalize it to, so the evaluator needs to handle it directly. The `query_macro_filter_parens` test exercises the exact panic path (`AND` containing a nested `OR` of `>` / `<` ops) and now runs on all drivers.

### DIFF
--- a/crates/toasty-driver-integration-suite/src/tests/crud_query_macro.rs
+++ b/crates/toasty-driver-integration-suite/src/tests/crud_query_macro.rs
@@ -136,11 +136,7 @@ pub async fn query_macro_filter_not(test: &mut Test) -> Result<()> {
     Ok(())
 }
 
-// Gated on `requires(sql)` until [#857] is fixed — `OR` of comparison ops
-// nested inside an `AND` panics in eval on the DynamoDB scan path.
-//
-// [#857]: https://github.com/tokio-rs/toasty/issues/857
-#[driver_test(id(ID), scenario(crate::scenarios::user_with_age), requires(sql))]
+#[driver_test(id(ID), scenario(crate::scenarios::user_with_age), requires(scan))]
 pub async fn query_macro_filter_parens(test: &mut Test) -> Result<()> {
     let mut db = setup(test).await;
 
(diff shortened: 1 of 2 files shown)


carllerche (approved):
Thanks!

PR #979, opened 2026-05-26
feat!: add increment, decrement, add, subtract update ops

carllerche:
## Summary

Add `stmt::increment()`, `stmt::decrement()`, `stmt::add(value)`, and `stmt::subtract(value)` assignment helpers for relative numeric updates (`col = col ± expr`). Atomic against the existing column value on every backend — SQL emits `col = col + :v`; DynamoDB uses the matching `SET col = col + :v` update expression. Closes a long-standing gap where the only way to do this was a racy round-trip read-modify-write.

```rust
user.update().balance(stmt::add(10)).exec(&mut db).await?;
user.update().login_count(stmt::increment()).exec(&mut db).await?;
```

Extends `BinaryOp` with `Add` / `Sub` (the enum's doc comment already anticipated arithmetic) and `Assignment` with `Add(Expr)` / `Subtract(Expr)`. `BinaryOp::commute()` now returns `Option<Self>` since `Sub` is not commutative.

## Type of change

- [x] Other: new public API. There is no pre-existing design doc; the surface mirrors existing assignment helpers (`stmt::push`, `stmt::pop`, `stmt::set`) so the convention is established.

## Checklist

- [x] `cargo test` passes (SQLite suite + DynamoDB suite both green)
- [x] Touches a public API
- [x] No `Driver` trait changes; driver-observable behavior is the new `Assignment::Add` / `Subtract` variants on update operations (covered by integration tests on every backend).

## Notes for reviewers

- DynamoDB driver: `ret` only includes non-`Set` assignments now, mirroring the engine's returning projection (which inlines `Set` values at plan time via `lower/returning.rs`). `Add` / `Subtract` go through `refresh_after_update` to fetch the post-update value from `UPDATED_NEW`.
- `BinaryOp::commute()` signature change is the breaking part — all four call sites are updated. Comparison-only paths (index match, in-subquery lifting, partition normalization) early-return / panic when commute is `None`; the constant-fold canonicalization swap is gated on `Some`.
- Simplify pass: `match-elimination-distributes-into-OR` rule is gated on `!op.is_arithmetic()` so arithmetic doesn't get distributed into a malformed boolean OR. Other comparison-specific rules were already gated on `is_eq()/is_ne()`.
- Integration tests in `crud_update_arithmetic.rs` cover increment, decrement, add, subtract (positive + negative), chained with a `Set` on another field, query-based update, and a driver-op shape check confirming the right `Assignment` variant flows through.


crates/toasty/src/stmt/assignment.rs
```diff
@@ -398,3 +402,78 @@ pub fn apply<T>(ops: impl IntoIterator<Item = Assignment<T>>) -> Assignment<T> {
...
+///
+/// # Examples
+///
+/// ```ignore
+/// user.update()
+///     .balance(stmt::add(10))
+///     .exec(&mut db)
+///     .await?;
+/// ```
+pub fn add<T>(value: impl IntoExpr<T>) -> Assignment<T> {
```
carllerche: Should `T: Numeric` here as well?

crates/toasty/src/stmt/assignment.rs
```diff
@@ -398,3 +402,78 @@ pub fn apply<T>(ops: impl IntoIterator<Item = Assignment<T>>) -> Assignment<T> {
...
+///
+/// # Examples
+///
+/// ```ignore
+/// user.update()
+///     .balance(stmt::subtract(5))
+///     .exec(&mut db)
+///     .await?;
+/// ```
+pub fn subtract<T>(value: impl IntoExpr<T>) -> Assignment<T> {
```
carllerche: Should `T: Numeric` here as well?

PR #980, opened 2026-05-26
feat(macros): add update! macro for concise field updates

carllerche:
## Summary

[Rendered](https://github.com/tokio-rs/toasty/blob/feat/update-macro/docs/dev/design/update-macro.md)

Adds the `update!` macro, a struct-literal-shaped form of the existing update builder. Lands the design doc and implementation together; updates the roadmap entry to point at the doc.

Supported field-entry shapes:

| Form | Expansion |
|---|---|
| `name: value` | `.name(value)` |
| `name` | `.name(name)` (Rust struct-literal shorthand) |
| `name.combinator(args)` | `.name(stmt::combinator(args))` — any `toasty::stmt::*` function |
| `field: { sub: val, ... }` | `stmt::apply([stmt::patch(<embedded root>.sub(), val), ...])` for embedded partial updates, recurses |
| `field: [{ ... }, expr, ...]` | `stmt::apply([stmt::insert(builder), expr, ...])` for has-many child inserts; bare lists of expressions stay `[expr, ...]` |

Bare-identifier instance targets work directly — `update!(user { ... })` expands to `user.update().…` and auto-borrows the same way the chain form does. The original binding stays owned after the macro. Query targets pass through and are consumed.

### Why no field-name validation

`create!`'s static-assertions story covers two things: required-field check and typo check. Update has no required fields. Typos surface through method resolution on the update builder — `update!(user { naem: "x" })` expands to `.naem("x")` and the compiler reports "no method named `naem`" at the macro call site. No custom validation infrastructure is needed.

### Why no operator sugar yet

The design doc covered `+=`, `-=`, `<<=` as sugar for `stmt::increment`, `stmt::decrement`, `stmt::push`/`stmt::insert`. Deferred to a follow-up because the engine has no `BinaryOp::Add`/`Sub`, no `Assignment::Increment` variant, and the SQL serializer can't reference the column being assigned. The macro's `.combinator(args)` shorthand reaches those builders once they exist. Tracked in the doc under `Out of scope`.

### What gets added in the derive

- `__macro_fields_root(&self) -> ModelFields<Model>` on every update builder — lets the macro recover the model's root fields struct from the bound builder, without needing to re-borrow the target value.
- `into_root(self) -> Self<Model>` on every fields struct — flips a path's `Origin` parameter to the field's own model so the path satisfies `stmt::patch`'s typed signature.

No new traits, no new helper functions in `codegen_support`. The only macro-support surface added to the derive is two `#[doc(hidden)]` inherent methods.

## Type of change

- [x] Other: combines a roadmap entry, a design doc, and the implementation in one PR — happy to split if maintainers prefer the separate-design-then-impl flow.

## Checklist

- [x] PR title uses Conventional Commits format
- [x] `cargo test -p tests --test sqlite` passes (1066 / 1066, including 20 new `update_macro` tests)
- [x] Touches a public API → design doc added at `docs/dev/design/update-macro.md`
- [x] No `Driver` trait changes

## Notes for reviewers

- The macro emits the existing public surface — no new engine variants, no new driver capability flags, no new `stmt::*` builders.
- Worth a focused look: the new `__macro_fields_root()` method on the update builder (in `model/expand/update.rs`) and the patch-path construction in `update/expand.rs` (uses `__macro_fields_root` then chains `.into_root()`).
- Bare `update!(user { ... })` works without the user writing `&mut user`; both shapes are accepted because the macro never binds the target — it evaluates the target once at the `.update()` call site, where autoborrow handles instance receivers.


docs/dev/design/update-macro.md
```diff
@@ -0,0 +1,493 @@
...
+    .exec(&mut db).await?;
+
+// Update — chain form
+user.update()
+    .name("Alice Smith")
+    .email("alice.smith@example.com")
+    .exec(&mut db).await?;
+```
+
+Two patterns make the chain form actively worse than a literal:
```
carllerche: This phrasing is negative. Rephrase it.

docs/dev/design/update-macro.md
```diff
@@ -0,0 +1,493 @@
...
+```
+
+This expands to `user.update().name("Alice Smith")` — the same call
+the chain form makes, with the same autoborrow behavior. `user` stays
+owned after the macro. Because the expansion is the existing
+instance-update path, optimistic-concurrency guards from a
+`#[version]` field still apply: the generated statement conditions on
+the version the instance was loaded with and increments it atomically.
+
+`update!(user { ... })` also works — `&mut user` is just one of
```
carllerche: This paragraph feels random. The call out to &mut user doesn't flow well. I suspect you should just remove it.

docs/dev/design/update-macro.md
```diff
@@ -0,0 +1,493 @@
...
+
+The most common form. A loaded model passes through unchanged:
+
+```rust
+let mut user = User::get_by_id(&mut db, &id).await?;
+
+update!(user { name: "Alice Smith" }).exec(&mut db).await?;
+```
+
+This expands to `user.update().name("Alice Smith")` — the same call
```
carllerche: This paragraph doesn't flow well. The autoborrow stuff is a non sequitur. Think hard about what the user trying to learn the update! Macro actually cares about. This is a guide not a reference.

docs/dev/design/update-macro.md
```diff
@@ -0,0 +1,493 @@
...
+update!(user { name: "Alice Smith" }).exec(&mut db).await?;
+```
+
+Today, the equivalent code is:
+
+```rust
+user.update().name("Alice Smith").exec(&mut db).await?;
+```
+
+The chain form is fine for one field. It gets noisy for several fields,
```
carllerche: This paragraph doesn't feel helpful to a user learning the update! Macro. I think there is some good content just the flow is awkward. Drop content not immediately useful in a guide and make sure the content flows well.

docs/dev/design/update-macro.md
```diff
@@ -0,0 +1,493 @@
...
+user.update()
+    .name("Alice Smith")
+    .email("alice.smith@example.com")
+    .exec(&mut db).await?;
+```
+
+Two patterns make the chain form actively worse than a literal:
+
+- **Multi-field updates on embedded types.** Partial updates lower to
+  `stmt::patch` + `stmt::apply`. Spelled out, that is a builder inside a
```
carllerche: This is not fact focused. There paragraph is meandering.

PR #983, opened 2026-05-27
fix: make starts_with case-sensitive on SQLite and MySQL

### LINKED ISSUE #936: `starts_with()` is case-insensitive on SQLite and MySQL
## Summary

`.starts_with(prefix)` is meant to be a case-sensitive prefix match with the
same behavior on every backend. It is case-sensitive on PostgreSQL (`^@`) and
DynamoDB (`begins_with`), but on SQLite and MySQL it lowers to `LIKE 'prefix%'`
and inherits `LIKE`'s case folding: case-insensitive for ASCII on SQLite, and
collation-dependent on MySQL. So `.starts_with("Al")` also matches `"alice"` on
SQLite.

Expected: only case-exact prefixes match, on every backend.
Actual: SQLite and MySQL also match prefixes that differ only by case.

## Reproducer

```rust
struct Item {
    #[key]
    id: i64,
    name: String,
}

// Seeded with "Alice" and "alice".
let items: Vec<Item> = Item::filter(Item::fields().name().starts_with("Al"))
    .exec(&mut db)
    .await?;

// PostgreSQL, DynamoDB: 1 row ("Alice")
// SQLite:               2 rows ("Alice", "alice")  <- wrong
```

## Database driver

SQLite, MySQL.

## Toasty version

commit 71271b63

## Additional context

- Root cause: `lower_expr_starts_with` rewrites `StartsWith` to `LIKE 'prefix%'`
  on SQL backends without a native prefix operator
  (`crates/toasty/src/engine/lower/expr_pattern.rs`), inheriting `LIKE`'s case
  behavior.
- Fix direction: emit a case-sensitive prefix match on SQLite and MySQL — e.g.
  SQLite `GLOB` (its wildcard syntax differs from `LIKE`) or a binary-collation
  comparison; MySQL a `_bin`/binary collation comparison. PostgreSQL `^@` and
  DynamoDB `begins_with` already match case-sensitively.
- Follow-up to #802, which settled `.like()`/`.ilike()` as pass-throughs (each
  keeps its backend's behavior) and `.starts_with()` as the operation Toasty
  implements with identical semantics everywhere. Making `.starts_with()`
  case-sensitive on all backends is the remaining work; the user-guide docs for
  `.starts_with()` were left unchanged until this lands.

Pritam-deb:
## Summary

Closes #936.

`starts_with()` was lowered to `LIKE 'prefix%'` for SQLite and MySQL, which is ASCII-case-insensitive on SQLite and collation-insensitive on MySQL. The intent is a byte-exact prefix match on every backend.

**Fix:** intercept `StartsWith` in `extract_params` (same pattern as `InList`/`AnyOp`/`AllOp`) and transform the prefix string before it is bound as a query parameter, then emit the appropriate operator in the serializer:

| Backend | Operator | Prefix transformation |
|---|---|---|
| SQLite / Turso | `col GLOB pattern` | `*`, `?`, `[` escaped as `[*]` `[?]` `[[]`; trailing `*` appended |
| MySQL | `BINARY col LIKE pattern ESCAPE '!'` | `!`, `%`, `_` escaped with `!`; trailing `%` appended |
| PostgreSQL | `col ^@` prefix | unchanged — already case-sensitive and handled natively |

Two new `Capability` flags (`glob_starts_with`, `binary_like_starts_with`) tell `extract_params` which transformation to apply. The old LIKE-lowering path in `lower/expr_pattern.rs` is removed; all SQL backends now set `native_starts_with: true`.

## Type of change

## Checklist

- [x] `cargo test` passes

## Notes for reviewers

The transformation happens at the `extract_params` stage — the earliest point where the raw prefix string is still visible before it becomes an opaque `Expr::Arg(n)`. The serializer therefore only needs to emit the right operator; it never sees the unescaped string.

`glob_starts_with` and `binary_like_starts_with` are mutually exclusive and both require `native_starts_with: true`; `Capability::validate()` enforces this.


Pritam-deb:
@carllerche the tests have been passing on my local, but here it fails.
Is there something i'm missing here?

carllerche:
Rerunning. It probably is related to the GitHub outage.

carllerche:
Thanks!

PR #997, opened 2026-05-29
refactor: remove compile-time field validation from create!

carllerche:
## Summary

The `create!` macro injected const-fn validation into **every** expansion — a per-model `__check_create_fields` const fn, `Model::CREATE_META` metadata, and monomorphization `__Check`/`assert_create_fields` blocks for scoped and nested creates. This removes all of it (including the `CreateMeta`/`CreateField`/`ValidateCreate` machinery) so we can measure how much it contributes to the `driver-integration-suite` crate's rustc memory use (a single build can hit ~7GB RSS).

Scoped-create safety is unchanged: the `CreateScope` bound on `scope_fields`/`create_in_scope` still rejects `via` and non-direct-relation scopes — only the error site moves (from the deleted `__force_check` to `scope_fields`). Missing required fields now surface at runtime as a DB constraint violation instead of a compile error, which the `create!` docs already describe.

## Type of change

## Checklist

- [x] `cargo test` passes

## Notes for reviewers

No public API of consequence is affected: `ValidateCreate` was `#[doc(hidden)]`, and `CreateMeta`/`CreateField`/`CREATE_META` were validation-only. Three UI `.stderr` files were regenerated: `create_invalid_field` loses the `E0080` "missing required field" panic (still fails on method-not-found), and the two scoped tests now derive the `CreateScope` failure from `scope_fields`.


Reinhard-Berger:
why not sitting this behind maby "toasty = { features = ["strict-create-checks"] }"

carllerche:
Seems plausible. You can give it a shot as a PR if you want.

PR #999, opened 2026-05-29
fix: deserialize a present Deferred<T> value as loaded

carllerche:
## Summary

Follow-up to #994. That PR's `Deserialize` read `null` as the unloaded state,
which collapsed a loaded `Deferred<Option<T>>` holding `None` into unloaded on
the way back in — a loaded field that silently became unloaded after a
serialize/deserialize round-trip, so `.get()` would then panic.

This deserializes any *present* value — including `null` — as loaded; the
unloaded state is carried by the field's absence via `#[serde(default)]`. The
encoding stays transparent (a loaded value is its bare inner `T`), matching how
the old `HasMany` relation wrapper handled serde.

Deferred fields are now expected to be skipped rather than serialized when
unloaded:

```rust
#[serde(skip_serializing_if = "Deferred::is_unloaded", default)]
notes: Deferred<Option<String>>,
```

Distinguishing unloaded from a loaded `None` without skipping would require a
non-transparent wrapper (the `lazy_slot` Value encoding does exactly that
internally); we keep the wire format transparent and document the skip pattern.

## Type of change

## Checklist

- [x] `cargo test` passes

## Notes for reviewers

The behavior change: a present `null` now means loaded, not unloaded. Without
the `skip_serializing_if` annotation an unloaded field still serializes to
`null` and reads back as loaded — that's the documented expectation, not a
round-trip we attempt to preserve.

### DIFF
--- a/crates/toasty/src/schema/deferred.rs
+++ b/crates/toasty/src/schema/deferred.rs
@@ -12,6 +12,22 @@ use std::fmt;
 /// value.
 ///
 /// `Deferred<Option<T>>` is supported when the column is nullable.
+///
+/// # Serde
+///
+/// With the `serde` feature a loaded `Deferred<T>` serializes transparently as
+/// its inner `T`, and any present value deserializes back to the loaded state.
+/// The unloaded state has no transparent encoding of its own — skip it on the
+/// way out and default it on the way in:
+///
+/// ```ignore
+/// #[serde(skip_serializing_if = "Deferred::is_unloaded", default)]
+/// notes: Deferred<Option<String>>,
+/// ```
+///
+/// Serializing an unloaded field without `skip_serializing_if` emits `null`,
+/// which does not round-trip (it reads back as loaded), so the annotation is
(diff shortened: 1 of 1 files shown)


joeydewaal:
Should we support deserialize with `Deferred`? I feel like this makes it easy to accidentally break your invariants. Especially when it's used with the relations. e.g:

```rust
#[derive(Model)]
struct User {
    ...
}

#[derive(Model, Deserialize)]
struct Todo {
    ...
    user_id: u64,

    #[belongs_to(key = user_id, references = id)]
    user: toasty::Deferred<User>,
}
```

When deserializing a `Todo` the `user_id` must be the same as the id of the user in the deferred user field. But we (toasty) can't really do that and it's up to the user check this. So this feels like a footgun.

Also we should probably ask for a use-case since I can't think of one. That's why I initially (https://github.com/tokio-rs/toasty/pull/143) didn't add support for deserialize in the now removed `Has{One,Many}` and `BelongsTo` structs.

PR #1001, opened 2026-05-29
feat: introduce Expr::Static for inline SQL literals, hook up LIMIT/OFFSET

MavenRain:
## Summary

Adds the `Expr::Static(Value)` AST variant proposed in the [Static SQL values design](https://github.com/tokio-rs/toasty/blob/main/docs/dev/design/static-sql-values.md).  `Expr::Static` is rendered inline as a SQL literal by the serializer and passes through `extract_params` untouched, in contrast to `Expr::Value` which is extracted as a bind parameter.

Switches the builder's `.limit(n)` / `.offset(n)` and cursor pagination's `page_size` from `Expr::Value` to `Expr::Static`, so `User::all().limit(10)` now renders as `LIMIT 10` inline instead of `LIMIT ?1` with `[I64(10)]` as a bind.  Plan caches in PostgreSQL, MySQL, and SQLite recognize more query shapes as the same.

Follow-up PRs hook up enum discriminants, `#[auto]` defaults, and variant-encoded `None` markers per the design's catalog.

Refs #237.

## Type of change

- [X] Implements a previously accepted [design](https://github.com/tokio-rs/toasty/blob/main/docs/dev/design/static-sql-values.md)
- [ ] Other (please explain):

## Checklist

- [X] `cargo test` passes


crates/toasty/src/stmt/paginate.rs
```diff
@@ -81,7 +81,7 @@ impl<M> Paginate<M> {
 
         let per_page = i64::try_from(per_page).expect("per_page exceeds i64::MAX");
         query.untyped.limit = Some(stmt::Limit::Cursor(stmt::LimitCursor {
-            page_size: stmt::Value::from(per_page).into(),
+            page_size: stmt::Expr::Static(stmt::Value::from(per_page)),
```
carllerche: I would not actually add any use of Static in this PR. I would save it for later. Caution needs to be used. I don't think per page is a good use case for static.

crates/toasty/src/stmt/query.rs
```diff
@@ -213,7 +213,7 @@ impl<T> Query<T> {
     pub fn limit(&mut self, n: usize) -> &mut Self {
         let n = i64::try_from(n).expect("limit exceeds i64::MAX");
         self.untyped.limit = Some(stmt::Limit::Offset(stmt::LimitOffset {
-            limit: stmt::Value::from(n).into(),
+            limit: stmt::Expr::Static(stmt::Value::from(n)),
```
carllerche: Same here, these aren't good applications of static.

carllerche (changes requested):
Some thoughts inline.

carllerche (approved):

Issue #863, opened 2026-05-10
Automatic retry on connection_lost for safe statements


carllerche:
### Problem

The connection-pool resilience design (#861) recovers the pool after a dead-connection event, but the caller still sees one `Error::connection_lost` per restart. For statements where Toasty can prove retry is safe, the engine could retry transparently and surface nothing.

Two cases are safe:

- **Read-only queries.** No side effects; retrying produces the same result.
- **The first statement of an explicit transaction.** The transaction has issued no other statements and no `COMMIT` has been sent, so the server-side rollback on connection drop leaves nothing to undo.

This is the deferred work called out in the design doc's "Out of scope" section under *Engine-level transparent retry of idempotent operations*.

### Proposed solution

On `Error::connection_lost` from `Connection::exec`, the engine retries on a fresh connection if either rule above holds. Retries are bounded (e.g., two attempts, matching Go's `database/sql`).

User-facing surface: the error stops reaching callers in the safe cases. Optionally a builder knob to disable transparent retry for callers who would rather see the error.

Out of scope for this proposal:

- Implicit single-statement writes (autocommit `INSERT`/`UPDATE`/`DELETE`). The implicit `COMMIT` is interleaved with the statement; a drop after server processing but before response is ambiguous. Users who want this safety can wrap the call in an explicit transaction and rely on the first-statement rule.
- Mid-transaction retry. After two or more statements in a transaction, retrying the failing statement alone produces inconsistent state; the user must restart the transaction.

### Alternatives considered

- **Per-query `#[idempotent]` annotation.** A model-level attribute marking a query as retryable. Heavier surface for less coverage — most read-only queries qualify automatically.
- **User-side wrappers.** Current behavior, kept available via `is_connection_lost()`. Works for users who know the predicate; silently broken for users who don't.
- **Driver-level "definitely-pre-send" classification.** A richer error variant from drivers indicating the request never left the client. Would let the engine retry autocommit writes safely, at the cost of per-driver classification work.

### Scope

Medium — engine-level change touching lower/plan/exec to know whether the operation is the first in a transaction, plus a small builder surface.

### Additional context

- Connection-pool resilience design: `docs/dev/design/connection-pool-resilience.md` (PR #861)
- Original pool-restart issue: #678
- First-pass recovery PR: #681

PR #1003, opened 2026-05-29
feat(engine): add statement effect classifier

MavenRain:
## Summary

Add `engine::effect::classify(&Statement) -> Effect` that returns `ReadOnly` for queries with no embedded mutation sub-statement and `Mutating` for everything else.  CTE-with-mutation queries (Postgres `WITH ins AS (INSERT ... RETURNING *) SELECT * FROM ins`) walk into the `WITH` clause and classify correctly as `Mutating`.

The classifier walks the AST via `stmt::Visit`, overriding `visit_expr_stmt` to flag any `Expr::Stmt(Insert | Update | Delete)` it encounters.  O(n) in tree size; no schema access.

First implementation PR for the [retry-safe-recovery design](https://github.com/tokio-rs/toasty/blob/main/docs/dev/design/retry-safe-recovery.md).  A follow-up PR will wire the pool's `connection_lost` handler to retry `ReadOnly` statements on a fresh connection.

Refs #863.

## Type of change

- [X] Implements a previously accepted [design](https://github.com/tokio-rs/toasty/blob/main/docs/dev/design/retry-safe-recovery.md)
- [ ] Other (please explain):

## Checklist

- [X] `cargo test` passes


crates/toasty/src/engine/effect.rs
```diff
@@ -0,0 +1,76 @@
...
+    /// Not safe to retry without further analysis.
+    Mutating,
+}
+
+/// Classify a statement's effect on database state.
+///
+/// O(n) in the size of the statement tree; no schema access.
+pub(crate) fn classify(stmt: &Statement) -> Effect {
+    match stmt {
+        Statement::Insert(_) | Statement::Update(_) | Statement::Delete(_) => Effect::Mutating,
```
carllerche: Nit, but I think you don't need to have the classification logic here since you also have it in visit_expr_stmt. You could probably create the walker and call visit_stmt with the initial statement. My guess is you can switch visit_expr_stmt to use visit_stmt? I might be missing something though.

carllerche (dismissed):

carllerche:
@codex review this again

carllerche (approved):

PR #1006, opened 2026-05-29
refactor!: remove the Register trait

carllerche:
## Summary

Registration used to traverse fields at the trait level, so both models and embeds implemented a shared `Register` trait to feed the registration API. Registration now discovers embeds transitively through model fields, so the shared trait is no longer needed.

- `id`/`schema`/`register` move onto `Model`; `id`/`schema` move onto `Embed`.
- Embed registration recurses through `Field::register` instead of a `register` method, and embeds are no longer submitted to inventory — they're reached only via the fields of the models that contain them.
- `Path` constructors bind on `Load` and recover the model id from `Load::ty()`.
- `models!` now accepts only `T: Model`.

**Breaking change:** the `Register` trait is removed, and `models!` no longer accepts embedded types directly (they're discovered through model fields).

Test fallout from the stricter `models!`: redundant embed args were stripped from `models!(...)` calls, the few tests that registered an embed standalone now wrap it in a throwaway model, and `discover` references its embed from a model.

## Type of change

## Checklist

- [x] `cargo test` passes

## Notes for reviewers

`Path`'s constructors now derive the `ModelId` from `Load::ty()` (a `model_id::<T>()` helper) rather than a dedicated trait method. Both `Model` and `Embed` return `Type::Model(id)` there, so this is the single point that unified the previously-shared `Register::id()`.


crates/toasty/src/stmt/path.rs
```diff
@@ -57,20 +70,20 @@ impl<T: Register> Path<T, T> {
     /// ```
     pub fn root() -> Self {
```
carllerche: Where is Path::root() used. Maybe we should add `fn path_root()` to the `Model` and `Embed` traits (with default implementations).

PR #1007, opened 2026-05-31
fix(sqlite): propagate errors instead of unwrapping in exec_sql

brendisurfs:
## Summary

Replaces two `unwrap` calls in the SQLite driver's `exec_sql` method with proper error propagation by mapping the driver-specific error to `toasty_core::Error::driver_operation_failed`. 
Previously, a failed `prepare_cached` or `query` call would panic, whereas now the error is returned to the caller. 

## Type of change

- [ ] Other (please explain):

## Checklist

- [x] `cargo test` passes

## Notes for reviewers

N/A


carllerche (approved):
ty

PR #1015, opened 2026-06-01
refactor!: rename RelationManyField/RelationOneField assoc type to Target

### LINKED ISSUE #1013: refactor: rename RelationManyField/RelationOneField `Model` assoc type to `Target`
## Problem

`RelationManyField` and `RelationOneField` expose the related model as an associated type named `Model`:

```rust
pub trait RelationManyField: Load<Output = Self> {
    type Model: Model;
    // …
}
```

PR #1012 added `ViaManyField`, whose analogous associated type is named `Target`. The relation traits and the via trait now disagree on the name for the same concept, and `<Vec<Todo> as RelationManyField>::Model` reads awkwardly (the bound and the name are both `Model`).

Affects generated code and `codegen_support`; these traits are part of toasty's surface but are referenced mostly by macro output.

## Proposed solution

Rename `RelationManyField::Model` and `RelationOneField::Model` to `Target`. Mechanical rename across the trait definitions, their impls, and the macro-generated references (`expand/fields.rs`, `expand/relation.rs`, `expand/schema.rs`, `expand/query.rs`, `expand/create.rs`, `expand/update.rs`).

## Alternatives considered

Rename `ViaManyField::Target` to `Model` instead. Discarded: a via target may be a scalar, so `Model` would be inaccurate there, whereas `Target` is correct for both. Aligning on `Target` is the accurate choice.

## Scope

## Additional context

Follows the naming introduced in #1012.

ab7tr4ct:
## Summary

`RelationManyField` and `RelationOneField` exposed the related model through an
associated type called `Model`. That name collided with the trait bound also
called `Model`, so `<Vec<Todo> as RelationManyField>::Model` ended up reading
awkwardly, and it didn't match `ViaManyField::Target` that #1012 introduced for
the same idea. This renames the associated type to `Target` on both traits, in
their impls, and in the macro-generated references.

Closes #1013.

## Type of change

- [x] Small / obvious change: internal cleanup, mechanical rename

## Checklist

- [x] `cargo test` passes

### DIFF
--- a/crates/toasty-macros/src/model/expand.rs
+++ b/crates/toasty-macros/src/model/expand.rs
@@ -393,8 +393,8 @@ impl Expand<'_> {
         let span = field_ident.span();
 
         quote_spanned! { span=>
-            #vis fn #field_ident(&self) -> <<#ty as #field_trait>::Model as #toasty::Model>::OneField<__Origin> {
-                <<<#ty as #field_trait>::Model as #toasty::Model>::OneField<__Origin>>::from_path(
+            #vis fn #field_ident(&self) -> <<#ty as #field_trait>::Target as #toasty::Model>::OneField<__Origin> {
+                <<<#ty as #field_trait>::Target as #toasty::Model>::OneField<__Origin>>::from_path(
                     self.path().chain(
                         <#model_ident as #schema_trait>::path_field(#field_offset)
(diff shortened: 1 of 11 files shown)


carllerche (approved):
ty :+1:

PR #1016, opened 2026-06-02
refactor(macros): drop dead RelationManyField::NULLABLE

### LINKED ISSUE #1014: refactor: investigate removing `NULLABLE` from RelationManyField
## Problem

`RelationManyField` declares a constant that is the same for every impl:

```rust
pub trait RelationManyField: Load<Output = Self> {
    /// A has-many is a collection; the collection itself is always present
    /// even when empty, so a has-many field is never nullable.
    const NULLABLE: bool = false;
    // …
}
```

The schema expansion reads `<#ty as RelationManyField>::NULLABLE` for the non-via has-many branch. After #1012, the via branch already emits `nullable = false` directly rather than going through the trait. A has-many collection is always present, so the constant never carries information.

Depends on #1012 landing (it introduced the hardcoded `false` on the via branch and split the has-many nullability handling).

## Proposed solution

Emit `false` for has-many nullability directly in `expand/schema.rs` for both branches and remove `RelationManyField::NULLABLE`.

## Alternatives considered

Leave the constant in place. Discarded: it is dead configuration — no impl overrides it, and it cannot meaningfully be `true`.

## Scope

## Additional context

Blocked on #1012.

ab7tr4ct:
## Summary

`RelationManyField::NULLABLE` was always `false` and no impl overrode it: a
has-many collection is always present, even when empty. After #1012 split the
has-many expansion, the via branch already emitted `false` directly. This does
the same for the non-via branch and removes the constant.

Closes #1014.

## Checklist

- [x] `cargo test` passes

### DIFF
--- a/crates/toasty-macros/src/model/expand/schema.rs
+++ b/crates/toasty-macros/src/model/expand/schema.rs
@@ -187,7 +187,8 @@ impl Expand<'_> {
                     } else {
                         let pair = expand_pair(toasty, quote!(#toasty::RelationManyField), ty, rel.pair.as_ref());
 
-                        nullable = quote!(<#ty as #toasty::RelationManyField>::NULLABLE);
+                        // A has-many collection is always present, never null.
+                        nullable = quote!(false);
                         deferred = quote!(<#ty as #toasty::RelationManyField>::DEFERRED);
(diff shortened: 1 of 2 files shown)


carllerche (approved):
ty

PR #1018, opened 2026-06-02
feat: support composite unique indices

### LINKED ISSUE #972: Capability to have Composite Unique Index
### Problem

I could find no way to create Composite Unique Index on Table level just like `#[index]` macro.

### Proposed solution

Creation of `unique` attribute just like `index` on table level.
```
#[unique(coa_id, combination_hash)]
pub struct AccountCombination {
    #[key]
    #[auto]
    id: uuid::Uuid,

    #[index]
    coa_id: uuid::Uuid,
    #[belongs_to(key = coa_id, references = id)]
    coa: toasty::BelongsTo<COA>,

    #[column(type = varchar(200))]
    #[unique]
    combination_hash: String,
}
```

### Alternatives considered

I have added in on local, happy to raise the PR after adding some tests.

### Scope

None

### Additional context

_No response_

carllerche:
## Summary

Implements composite unique indices (#972). A struct-level `#[unique(field_a, field_b)]` declares a unique index spanning multiple columns, mirroring `#[index(...)]` (simple and partition/local modes, plus `name = "..."`) but enforcing uniqueness across the listed fields. Previously only single-field `#[unique]` was available.

```rust
#[unique(coa_id, combination_hash)]
struct AccountCombination {
    #[key] #[auto] id: u64,
    coa_id: i64,
    combination_hash: String,
}
```

The `unique` flag already threaded through the schema layers, so SQL backends emit `CREATE UNIQUE INDEX (...)` with no engine changes; query-method codegen is uniqueness-agnostic and generates the usual `get_by_*` / `filter_by_*` for each index prefix.

DynamoDB backs each unique index with a dedicated single-column index table guarded by `attribute_not_exists`, which cannot enforce a composite constraint. `create_table` now rejects multi-column unique indices with `unsupported_feature` (validated before any table is created) instead of panicking on the old `assert_eq!(1, …)`. Single-column `#[unique(field)]` is unchanged there.

Closes #972

## Type of change

- [x] Other: additive feature implementing accepted request #972; mirrors the existing `#[index(...)]` attribute, documented in the user guide.

## Checklist

- [x] `cargo test` passes — SQLite full suite (1108) and DynamoDB full suite against local DDB, plus trybuild UI tests

## Notes for reviewers

- No design doc: the feature mirrors the existing `#[index(...)]` / `#[unique]` attributes and is documented in the user guide (`indexes-and-unique-constraints.md`, `dynamodb.md`). Issue #972 serves as the design.
- DynamoDB composite-unique support is deliberately deferred. The index-table mechanism (and the insert/update/delete sync) is single-column; full support would need a composite-key index table plus rewritten write synchronization. The clean `unsupported_feature` rejection keeps single-column unique working while making the gap explicit.
- New tests in `index_composite.rs`: `composite_unique_index_enforced` (SQL — duplicate combination rejected, partial overlaps allowed) and `composite_unique_index_unsupported_on_dynamodb` (DDB rejection).

### DIFF
--- a/crates/toasty-driver-dynamodb/src/op/create_table.rs
+++ b/crates/toasty-driver-dynamodb/src/op/create_table.rs
@@ -57,7 +57,26 @@ impl Connection {
         let mut gsis = vec![];
 
         for index in &table.indices {
-            if index.primary_key || index.unique {
+            if index.primary_key {
+                continue;
+            }
+
+            if index.unique {
+                // Unique indices are materialized as separate tables below. Each
+                // index table is keyed on a single column and enforced with an
+                // `attribute_not_exists` condition, so composite (multi-column)
+                // unique indices are not supported. Reject them here, before any
+                // table is created, rather than partway through. SQL backends
+                // support them via `CREATE UNIQUE INDEX`.
+                if index.columns.len() != 1 {
+                    return Err(toasty_core::Error::unsupported_feature(format!(
+                        "DynamoDB does not support composite unique indices; \
+                         index '{}' spans {} columns",
+                        index.name,
+                        index.columns.len()
+                    )));
+                }
+
                 continue;
             }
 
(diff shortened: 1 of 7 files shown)


carllerche:
Fixed the MySQL CI failure in 74852ceb2. The auto-generated index name (`index_<table>_by_<col>_and_<col>`) plus the integration harness's per-test table prefix exceeded MySQL's 64-char identifier limit. Shortened the test's model/field names so it fits while keeping auto-naming coverage; verified `composite_unique_index_enforced` now passes against a local MySQL.

This is a pre-existing fragility (any long-enough composite `#[index]` hits it), not specific to this feature, so I kept the fix scoped to the test and filed #1020 for general backend-aware identifier-length handling.

PR #1019, opened 2026-06-02
feat: support Option<Embed> as a model field

### LINKED ISSUE #800: feat: support Option<EmbeddedType> as a model field
## Problem

A model field typed `Option<MyEmbed>` (where `MyEmbed: #[derive(toasty::Embed)]`) does not compile through the schema layer. The build of the schema rejects the column type with:

```
unsupported feature: type Model(ModelId(...)) is not supported by this database
```

This came up while writing tests for #799 (extending `#[deferred]` to embedded types). The user-facing patterns the design covers — `Deferred<Option<EmbedStruct>>`, `Deferred<Option<EmbedEnum>>` — can't be tested or used because the underlying `Option<Embed>` shape isn't supported.

Reproducer:

```rust
#[derive(toasty::Embed)]
struct Metadata {
    author: String,
}

#[derive(toasty::Model)]
struct Document {
    #[key] #[auto] id: u64,
    metadata: Option<Metadata>,   // panics during setup
}
```

## Proposed solution

Lift `Option<Embed>` to a first-class field shape:

- Schema: a nullable embedded field flattens to nullable columns.
- Lowering / loading: a row where the embedded primary fields are all NULL decodes to `None`. (Choice of discriminant — implicit "all-null" vs. an explicit marker column — is part of the design.)
- Macros: `<Option<T> as Field>::Path<Origin>` should reach the embed's path struct so `Document::fields().metadata().author()` works the same as for a non-optional embed.

## Alternatives considered

- Require users to wrap a single column they treat as the discriminator. Loses ergonomics; not a real alternative.
- Defer to `BelongsTo<Option<T>>`. Different semantics (separate table, not flattened).

## Scope

Medium — touches the schema layer, lowering, and the embed-codegen path in `toasty-macros`.

## Additional context

- #799 — extends `#[deferred]` to embedded types. The implementation is structured so the deferred-mask walk descends into embed records; once `Option<Embed>` lands, `Deferred<Option<EmbedType>>` should compose without a separate code path. The fix here should consider that interaction — at minimum, verify that `Deferred<Option<Embed>>` round-trips for the optional + loaded-None and optional + loaded-Some cases (the same wire-encoding decision tracked in #797 applies).
- #797 — discusses the nullable-inner sentinel for `BelongsTo` / `Deferred`; relevant to how `Option<Embed>` distinguishes "loaded as None" from row-level absence in include subqueries.

carllerche:
## Summary

Adds `Option<StructEmbed>` model fields (issue #800).

The issue's premise was stale: the "type `Model(...)` is not supported" error
predated #926, which made `Option<T>::field_ty` delegate to the inner type, so
`Option<Embed>` already resolved to an `Embedded` app field. What was actually
broken was that the flattened columns were `NOT NULL`, inserting `None` panicked
in the simplifier (`Project(Null, ..)` → `Value::entry` `todo!()`), and a `None`
row never decoded back to `None`.

A nullable struct embed is a two-state value (`None` = every flattened column
NULL, vs `Some(record)`) — structurally the same as the existing enum embed.
The fix reuses that machinery, so **no macro / `Field`-trait changes** are
needed (the accessor already yields `Path<_, Option<Embed>>`):

- **`core`** — `MapField.in_nullable_embed` forces every flattened leaf column
  nullable; an encode null-guard makes a `None` write all-NULL; a decode
  sentinel `Match(all-columns-NULL → null, else record)` (in `table_to_model`
  and the INSERT…RETURNING default) decodes an all-NULL row to `None`.
- **`engine`** — `IS NULL` distributes over `Record`/`Match` so
  `.is_none()`/`.is_some()` reduce to per-column predicates (neither SQL nor
  DynamoDB can carry a `Match` in a predicate).

Two latent precedence bugs the new expressions exposed are also fixed:

- **`sql`** — parenthesize `OR` operands nested inside `AND`; without it
  `a AND (b OR c)` serialized as `a AND b OR c`.
- **`engine`** — prune dead `AND(.., NULL)` match-elimination terms. SQL
  tolerated `… AND NULL`, but DynamoDB rejected the bare value placeholder in a
  `FilterExpression`.

Covers create/read `Some`+`None`, set-whole-value update, `.is_none()` /
`.is_some()`, and whole-value `.eq(Some(..))`, on SQLite, PostgreSQL, and
DynamoDB.

**Not** `Closes #800`: the issue also lists nested-field path access
(`Document::fields().metadata().author()`) as a goal. That conflicts with
nullable-scalar path typing (`Path<_, Option<U>>` carries `is_none`/`is_some`)
and would need specialization or a `Field` path-machinery redesign, so it's
deferred — along with `Option<enum embed>` and partial within-embed updates.

## Type of change

- [x] Other: implements the maintainer-authored proposed solution in #800
      (struct-embed subset; nested path access deferred).

## Checklist

- [x] `cargo test` passes — SQLite 2944, PostgreSQL 1150, DynamoDB 534, all green

## Notes for reviewers

- The design is captured in #800; no separate `docs/dev/design/` doc was added.
- The two precedence fixes are independent latent bugs, only surfaced here
  because the nullable-embed sentinel is the first expression to nest `OR`
  inside `AND` / produce an `AND(.., NULL)` filter term. Worth a close look —
  they change SQL/DDB filter serialization for any future expression of that
  shape.
- `.eq(None)` is intentionally not tested: `expr <op> null` folds to constant
  `null` for every field (pre-existing), so `.is_none()` is the idiom.

### DIFF
--- a/crates/toasty-core/src/schema/builder/table.rs
+++ b/crates/toasty-core/src/schema/builder/table.rs
@@ -77,6 +77,13 @@ struct MapField<'a, 'b> {
     /// since only the active variant's columns are populated.
     in_enum_variant: bool,
 
+    /// When true, columns are created nullable regardless of the field's own
+    /// nullability. Set while processing the fields of a nullable embed
+    /// (`Option<Embed>`), since the entire embed may be absent (all columns
+    /// NULL). Propagated into nested embeds so every flattened leaf column of
+    /// an optional embed is nullable.
+    in_nullable_embed: bool,
+
     /// Base expression for the current nesting level.
     ///
     /// `None` at the top level. `Expr::Project(source_ref, proj)` at any nested
@@ -488,6 +495,15 @@ impl BuildMapping<'_> {
                         .collect::<Result<_>>()?;
                     stmt::Expr::record(exprs)
                 };
+                // A nullable embed's default-returning expression must also
+                // collapse to `Null` when the embed is absent, so an
+                // INSERT...RETURNING of a `None` echoes back `None` rather than
+                // `Some` of all-NULL fields.
+                let expr = if field.nullable {
+                    self.wrap_nullable_embed_record(s, expr)
+                } else {
+                    expr
+                };
                 s.default_returning = expr.clone();
                 Ok(expr)
             }
@@ -719,6 +735,7 @@ impl BuildMapping<'_> {
         &self,
         model: &app::EmbeddedStruct,
         mapping: &mapping::FieldStruct,
+        nullable: bool,
     ) -> Result<stmt::Expr> {
         let exprs: Vec<stmt::Expr> = model
             .fields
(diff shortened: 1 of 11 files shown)


carllerche:
#1021 is a better attempt at this.

PR #1022, opened 2026-06-03
fix(engine): increment #[version] field on query-based updates

### LINKED ISSUE #982: Query-based updates do not increment #[version] field
### Summary

Query-based updates on a model with `#[version]` do not increment the version column. Instance updates do.

### Reproducer

```rust
use toasty::Model;

struct Item {
    #[key] #[auto] id: uuid::Uuid,
    name: String,
    #[version] version: u64,
}

let mut item = toasty::create!(Item { name: "a" }).exec(&mut db).await?;
assert_eq!(item.version, 1);

Item::filter_by_id(item.id)
    .update()
    .name("b")
    .exec(&mut db)
    .await?;

let reloaded = Item::filter_by_id(item.id).get(&mut db).await?;
assert_eq!(reloaded.name, "b");
assert_eq!(reloaded.version, 2); // fails: version is still 1
```

The existing integration test `query_update_multi_key_works` in `crates/toasty-driver-integration-suite/src/tests/field_version.rs` documents the current behavior in its docstring.

The macro emits the version assignment + condition only on the instance-rooted `Model::update(&mut self)` method (`crates/toasty-macros/src/model/expand/model.rs`, calling `expand_version_update_stmts` in `update.rs`). The query-rooted `From<#query>` impl for the update builder applies update defaults but never touches the version field.

A guard is not appropriate for query-based updates (the update is atomic at the DB level), but the version should still increment.

### Database driver

DynamoDB. SQL drivers do not yet implement `#[version]` OCC, per `docs/guide/src/concurrency-control.md`.

### Toasty version

Commit `62f7765d` on `main`.

### Additional context

Implementation depends on atomic `col = col + :v` assignment, added by #979 (`feat!: add increment, decrement, add, subtract update ops`). Once #979 lands, the macro can emit `Assignment::Add(1)` on the version field in the query-rooted update path.

carllerche:
## Summary

Query-based updates (`Model::filter_by_id(id).update()…`) left the `#[version]`
counter untouched, so a concurrent read-modify-write — an instance update
conditioned on a stale version — could overwrite a query update's write without
tripping the OCC check.

The engine now injects `version = version + 1` during update lowering for
versionable models whose version field isn't already assigned: instance updates
set it (with the matching `version == <last read>` condition) in generated
code; query updates did not. Injecting after the `Changed` returning projection
is built keeps the engine-managed counter out of the returning clause, so a
relative increment never forces a per-row read-back the multi-row DynamoDB
transact path cannot serve.

Closes #982.

## Type of change

## Checklist

- [x] `cargo test` passes

## Notes for reviewers

The increment lives in the engine (`engine/lower.rs`), not the macro, so it
applies uniformly however the update statement is built. No `Driver` trait
change — the DynamoDB driver already handles `Assignment::Add`. The
driver-observable change is that query updates on a versioned model now carry a
`version + 1` assignment.

Coverage (DynamoDB only; SQL drivers don't implement `#[version]` OCC):
single-key, multi-key `transact_write_items`, the unique-index path, a
`Version(u64)` newtype query update, and a test that a query update invalidates
a concurrent stale instance update.


carllerche:
@codex review this PR again

PR #1023, opened 2026-06-03
fix(core): truncate auto-generated index names that exceed the backend limit

### LINKED ISSUE #1020: Auto-generated index names can exceed MySQL's 64-char identifier limit
## Summary

Toasty derives index names as `index_<table>_by_<col>` (joining further columns with `_and_<col>`). For a composite `#[index(...)]` or `#[unique(...)]` on a model with long-ish table and column names, the generated name exceeds MySQL's 64-character identifier limit and `CREATE INDEX` fails with `ERROR 1059 (42000)`. PostgreSQL silently truncates at 63 bytes (latent collision risk if two indexes share a 63-char prefix); SQLite and DynamoDB are unaffected.

The integration test harness prepends a per-test table prefix (`test_{pid}_{counter}_`), which makes this easier to hit in tests — it surfaced while adding composite-unique-index tests.

## Reproducer

```rust
// index name: index_organization_memberships_by_organization_id_and_member_user_id (69 chars)
#[index(organization_id, member_user_id)]
struct OrganizationMembership {
    #[key] #[auto] id: u64,
    organization_id: i64,
    member_user_id: i64,
}
// setup against MySQL -> ERROR 1059 (42000): Identifier name '...' is too long
```

## Database driver

MySQL (hard error). PostgreSQL truncates silently at 63 bytes. SQLite / DynamoDB unaffected.

## Toasty version

main @ c05853373

## Additional context

`update_index_names` in `crates/toasty-core/src/schema/builder/table.rs` generates these names and already has the driver `Capability` in scope, and each `Db` is bound to one driver — so a backend-aware cap is feasible at build time.

Proposed fix:
- Add `max_identifier_length: Option<usize>` to `Capability` (MySQL `Some(64)`, PostgreSQL `Some(63)`, SQLite/DynamoDB `None`).
- When a generated name exceeds the limit, truncate and append a short suffix derived from a **stable** hash of the full intended name (not `DefaultHasher`, which is unstable across builds and would churn migration diffs). User-provided `#[index(name = "...")]` is left untouched.

Affects both Toasty users (any sufficiently long composite index) and driver implementors (the limit is per-backend).

Pritam-deb:
## Summary

Fixes #1020.

`update_index_names` generates names of the form `index_<table>_by_<col>(_and_<col>)*`. For models with long-ish table and column names — especially under the test harness, which prepends a `test_{pid}_{counter}_` prefix — these names exceed MySQL's hard 64-char limit (`ERROR 1059`) and PostgreSQL's 63-char soft limit (silent truncation with latent collision risk).

**Changes:**

- Add `Capability::max_identifier_length: Option<usize>` — `Some(64)` for MySQL, `Some(63)` for PostgreSQL, `None` for SQLite and DynamoDB.
- In `update_index_names`, when a generated name exceeds the limit, truncate to `limit - 5` bytes and append `_XXXX`: four hex digits from an FNV-1a hash of the *full* pre-truncation name. The hash is stable across builds (unlike `DefaultHasher`) so migration snapshots don't churn. Two indexes that share a long prefix get different suffixes.
- User-supplied `#[index(name = "...")]` names are never touched.

**Test:** `index_long_name_is_truncated` creates the exact model from the issue reproducer (`OrganizationMembership` with a two-column composite index) and asserts the index name respects the backend limit.

## Type of change

## Checklist

- [x] `cargo test` passes

## Notes for reviewers

FNV-1a is inlined (~5 lines) to avoid a new dependency. Only the low 16 bits of the hash are used (`& 0xFFFF`) — that's 65 536 buckets, which is more than enough to distinguish the handful of long indexes a schema is likely to have.

### DIFF
--- a/crates/toasty-core/src/driver/capability.rs
+++ b/crates/toasty-core/src/driver/capability.rs
@@ -67,6 +67,19 @@ pub struct Capability {
     /// `Integer(4)`.
     pub max_auto_increment_integer_width: Option<u8>,
 
+    /// Maximum byte length for a database identifier (table name, index name,
+    /// column name, etc.).
+    ///
+    /// When `Some(n)`, auto-generated index names that exceed `n` bytes are
+    /// truncated and a short stable hash suffix is appended so names remain
+    /// unique and deterministic across builds. User-supplied `#[index(name =
+    /// "...")]` names are left untouched.
+    ///
+    /// - MySQL: `Some(64)` — hard error on longer names
+    /// - PostgreSQL: `Some(63)` — silently truncates, risking collisions
+    /// - SQLite / DynamoDB: `None` — no enforced limit
+    pub max_identifier_length: Option<usize>,
+
     /// Whether the database supports `VARCHAR(n)` column types natively.
     ///
     /// Must be consistent with [`StorageTypes::varchar`]: when `true`,
@@ -529,6 +542,7 @@ impl Capability {
         max_auto_increment_integer_width: Some(4),
         bigdecimal_implemented: false,
         bool_key_type: true,
+        max_identifier_length: None,
 
         native_varchar: true,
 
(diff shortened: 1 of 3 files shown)


carllerche (approved):
thanks :+1:

PR #1028, opened 2026-06-04
feat: add #[document] storage for embedded types with nested-path filtering

carllerche:
## Summary

Adds the `#[document]` field attribute, which stores an `#[derive(Embed)]` value
as a single JSON document column (`jsonb` on PostgreSQL, JSON1 text on SQLite)
instead of column-expanding it into one column per leaf. Two shapes are
supported:

- A `Vec<T>` of embedded structs — stored as a JSON array of objects.
- A bare struct embed — stored as one JSON object — which keeps the full path
  query API. `Account::fields().profile().address().city().eq("Seattle")`
  lowers to a JSON extraction in the `WHERE` clause, so a document-stored embed
  filters the same way a column-expanded one does: equality, range, and
  `is_none`, at any nesting depth.

Whole-value create / read / update work for both shapes. MySQL and DynamoDB
reject `#[document]` at schema build (gated by the `document_collections`
capability). This is the first slice of the larger document-fields design;
maps, sets, sub-document containment, per-element mutation, indexes, and the
remaining backends are follow-ups.

## Type of change

- [x] Implements a previously accepted design:
      [`docs/dev/design/document-fields.md`](https://github.com/tokio-rs/toasty/blob/worktree-indexed-wibbling-gizmo/docs/dev/design/document-fields.md)

## Checklist

- [x] `cargo test` passes (SQLite locally; PostgreSQL / MySQL / DynamoDB run in CI)
- [x] Touches driver-observable behavior → the design doc covers it

## Notes for reviewers

- **How it works.** A `#[document]` embed resolves to `stmt::Type::Document`; the
  engine builds a named `Value::Object` at the driver boundary (where the
  column's `TypeDocument` supplies field names) and collapses it back to a
  positional `Value::Record` on load. Nested-path filtering adds an
  `ExprFunc::JsonExtract` node, produced by the simplifier when a projection
  lands on a document-typed column and serialized per dialect (`col->'a'->>'b'`
  on PostgreSQL, `json_extract(col, '$.a.b')` on SQLite / MySQL).
- **Provenance.** The `Vec<Embed>` storage foundation (the
  "document-storage foundation …" commit) is adopted from the in-flight branch
  `worktree-splendid-marinating-corbato`, reconciled onto current `main`. The
  bare-embed storage and all nested-path filtering are new in this PR.
- **Design doc.** This branch also carries a rewrite of
  `docs/dev/design/document-fields.md` to a remaining-work framing. It describes
  the full surface; this PR implements the first slice.
- **Trybuild.** `relation_has_one_requires_attr.stderr` is refreshed for the
  unified Field/Document setter codegen — still a compile-fail with the same
  guidance. The pinned toolchain (1.95.0) matches CI.
- Verified end-to-end on SQLite; the PostgreSQL paths exercise the same codec
  and run in CI.


Caisin:
# Proposal: Dynamic JSON document fields with direct point lookup

## Summary

Add support for using `serde_json::Value` as a document-capable field, with a small path lookup API for filtering JSON leaves.

The target API is:

```rust
#[derive(toasty::Model)]
struct Item {
    #[key]
    id: i64,

    #[document]
    data: serde_json::Value,
}

let items = Item::filter_by_data(|data| {
    data.point::<String>("a/b/c").eq("abc")
});
```

This proposal does not ask for full typed `Embed` behavior for `serde_json::Value`. It only asks for dynamic JSON storage and direct leaf comparison by path.

## Problem

Toasty already supports `#[document]` for `#[derive(Embed)]` structs and `Vec<Embed>` collections. That works well when the JSON structure is known at compile time.

Some fields need dynamic JSON:

```rust
serde_json::Value
```

Examples:

- metadata received from external systems
- user-defined attributes
- API payload snapshots
- flexible configuration blobs
- audit context with changing keys

Today the closest supported field type is:

```rust
data: toasty::Json<serde_json::Value>
```

That stores the value as JSON text, but it is an opaque serialized field. It does not use Toasty's document field path, and it does not allow filtering by a JSON leaf.

`serde_json::Value` should not implement `Embed`. `Embed` represents a statically known schema with generated field paths. `serde_json::Value` has no fixed field list.

## Proposed API

Add document support for `serde_json::Value`:

```rust
#[derive(toasty::Model)]
struct Item {
    #[key]
    id: i64,

    #[document]
    data: serde_json::Value,
}
```

Add a path lookup method on the generated field path:

```rust
Item::fields()
    .data()
    .point::<String>("a/b/c")
    .eq("abc")
```

Other examples:

```rust
Item::fields().data().point::<i64>("count").eq(3)
Item::fields().data().point::<bool>("flags/enabled").eq(true)
Item::fields().data().point::<String>("profile/name").ne("Alice")
```

The type parameter gives Toasty the expected leaf type. This keeps the API small and avoids adding a dynamic JSON type system to the query builder.

## Path syntax

The public API can start with slash-separated paths:

```rust
point::<String>("a/b/c")
```

The implementation should parse the string into path segments:

```rust
["a", "b", "c"]
```

If keys may contain `/` or `~`, the API should either:

1. use JSON Pointer escaping:
   - `~1` means `/`
   - `~0` means `~`

or:

2. add a segment-based form later:

```rust
point_segments::<String>(["a", "b/c", "d"])
```

The first version can document the supported escaping rules and reject invalid paths.

## Intended scope

The first implementation only needs to support:

- `#[document] data: serde_json::Value`
- whole-field create / load / update
- leaf equality and inequality through `point::<T>(...)`
- scalar leaf types:
  - `String`
  - integer types
  - floating-point types
  - `bool`
  - `uuid::Uuid`, if existing JSON extraction casts support it

The first implementation does not need to support:

- implementing `Embed` for `serde_json::Value`
- generated typed subfields
- arbitrary chained `.path().path()` builders
- containment queries
- array element predicates
- partial document updates
- indexing dynamic JSON paths
- schema validation for dynamic JSON

## Implementation pattern

### 1. Add a dynamic JSON statement type

Add a statement-level type for dynamic JSON documents, for example:

```rust
stmt::Type::Json
```

or:

```rust
stmt::Type::Document
```

This type represents a self-describing JSON value. It is different from:

```rust
stmt::Type::Model(embed_id)
```

`Type::Model(embed_id)` means a document with a known embedded schema. The new type means a document with dynamic keys and dynamic values.

### 2. Map the new type to document storage

Update database type mapping so the new type maps to document storage:

```rust
db::Type::Document { binary: true }
```

This should match existing document storage behavior:

- PostgreSQL: `jsonb`
- MySQL: `JSON`
- SQLite: JSON text / JSON1-compatible storage
- DynamoDB: Map/List/scalar attribute values, if supported

### 3. Mark dynamic JSON columns as document columns

Update document detection so both static and dynamic document columns are recognized.

Current static document cases are conceptually:

```rust
Type::Model(_)
Type::List(Model(_))
```

Add the dynamic JSON type:

```rust
Type::Json
```

This lets the existing document lowering and driver paths recognize the column as document-backed.

### 4. Implement Toasty traits for `serde_json::Value`

Inside Toasty, implement the required local traits for `serde_json::Value`:

- `Load`
- `Field`
- `Document`
- `IntoExpr`
- assignment support

The `Field` implementation should expose a path type with `point::<T>(...)`.

The `Document` implementation should allow:

```rust
#[document]
data: serde_json::Value
```

at compile time.

### 5. Convert between `serde_json::Value` and `stmt::Value`

Add conversion helpers:

```rust
serde_json::Value -> stmt::Value
stmt::Value -> serde_json::Value
```

Suggested mapping:

| JSON value | Toasty value |
|---|---|
| `null` | `Value::Null` |
| `bool` | `Value::Bool` |
| string | `Value::String` |
| number | `Value::I64`, `Value::U64`, or `Value::F64` |
| array | `Value::List` |
| object | `Value::Object` |

Number handling needs a clear rule. A conservative rule is:

- use `I64` when the JSON number fits `i64`
- use `U64` when it fits `u64`
- otherwise use `F64`
- reject numbers that cannot be represented by Toasty's value model

### 6. Reuse `FuncJsonExtract`

Toasty already has:

```rust
FuncJsonExtract {
    base,
    path: Vec<String>,
    ty,
}
```

The `point::<T>("a/b/c")` method can build this expression directly.

For example:

```rust
data.point::<String>("a/b/c")
```

can produce:

```rust
FuncJsonExtract {
    base: data_column_expr,
    path: vec!["a", "b", "c"],
    ty: stmt::Type::String,
}
```

Then existing SQL serialization can render the expression as:

- PostgreSQL: `data->'a'->'b'->>'c'`
- SQLite: `json_extract(data, '$.a.b.c')`
- MySQL: `JSON_EXTRACT(data, '$.a.b.c')` with existing unquote/cast handling

### 7. Add a typed dynamic JSON path expression

The public method can return a typed expression wrapper:

```rust
fn point<T>(&self, path: impl AsRef<str>) -> Expr<T>
where
    T: JsonLeaf;
```

`JsonLeaf` can map Rust leaf types to `stmt::Type`.

This keeps comparison type checking intact:

```rust
data.point::<String>("a/b/c").eq("abc")
data.point::<i64>("a/b/c").eq(123)
```

### 8. Keep static `Embed` behavior unchanged

Existing `#[derive(Embed)]` document fields should keep their current behavior.

Static embed documents still use schema-directed paths:

```rust
user.preferences().theme().eq("dark")
```

Dynamic JSON fields use string paths:

```rust
item.data().point::<String>("preferences/theme").eq("dark")
```

These two APIs solve different problems.

## Suggested first test cases

### Create and get

```rust
struct Item {
    #[key]
    id: i64,

    #[document]
    data: serde_json::Value,
}

let value = serde_json::json!({
    "a": {
        "b": {
            "c": "abc"
        }
    }
});

let item = Item::create().data(value.clone()).exec(&mut db).await?;
let loaded = Item::get_by_id(&mut db, &item.id).await?;

assert_eq!(loaded.data, value);
```

### Equality filter

```rust
let item = Item::filter(
    Item::fields().data().point::<String>("a/b/c").eq("abc")
)
.first(&mut db)
.await?;
```

### Numeric filter

```rust
let item = Item::filter(
    Item::fields().data().point::<i64>("count").eq(3)
)
.first(&mut db)
.await?;
```

### Boolean filter

```rust
let item = Item::filter(
    Item::fields().data().point::<bool>("enabled").eq(true)
)
.first(&mut db)
.await?;
```

## Why this is smaller than full dynamic JSON query support

A full dynamic JSON API would need chained path builders, dynamic leaf typing, array indexing, containment, mutation, and indexing support.

This proposal avoids that scope.

The `point::<T>(...)` API gives Toasty enough information to generate one typed JSON extraction expression. It reuses the existing `FuncJsonExtract` representation and existing backend serializers.

## Open questions

1. Should the path syntax be `"a/b/c"` or JSON Pointer `"/a/b/c"`?
2. Should missing paths compare as SQL `NULL`?
3. Should `point::<T>()` support `Option<T>` for missing or null values?
4. Should the first implementation support DynamoDB, or start with SQL drivers?
5. How should JSON numbers outside `i64` / `u64` / `f64` be handled?

## Requested support

Please consider adding `serde_json::Value` as a special document-capable field type, with a small `point::<T>(...)` leaf lookup API.

This would let Toasty support dynamic JSON documents without treating them as `Embed` types and without adding a large dynamic JSON query surface in the first version.

PR #1031, opened 2026-06-07
refactor(core): infer storage type directly from Value

carllerche:
## Summary

Raw SQL bind inference (`toasty::sql`) typed each `Param::Infer` value by calling
`Value::infer_ty` and then `db::Type::from_app`. That builds the richer
`stmt::Type` — record recursion, list `TypeUnion` construction and simplification —
only for `from_app` to immediately collapse it back to a `db::Type`.

This adds `Value::infer_db_ty`, which maps each value variant straight to its
storage type, and switches raw SQL bind inference to it. No behavior change:

- Scalars resolve through the driver's `StorageTypes` defaults (same source `from_app` uses).
- Lists map to their (uniform, non-`NULL`) element type.
- Values whose type can't be inferred from the value alone (`NULL`, records,
  empty / all-`NULL` / mixed lists) still error with guidance to use `bind_typed`.

## Type of change

- [x] Small / obvious change — internal cleanup

## Checklist

- [x] `cargo test` passes

Adds a public `Value::infer_db_ty` method, but it introduces no user-observable
behavior — it mirrors the existing public `Value::infer_ty` and the raw SQL bind
path it replaces behaves identically. No design doc needed; no `Driver`-trait or
driver-observable change.

## Notes for reviewers

The list edge cases are the one place to check: the previous path errored on
empty / all-`NULL` / heterogeneous lists incidentally (via `from_app` after
`TypeUnion` collapse); `infer_db_ty` errors on them explicitly. The behavior is
the same, just reached directly.


crates/toasty-core/src/stmt/value.rs
```diff
@@ -407,6 +407,101 @@ impl Value {
...
+    /// ```
+    pub fn infer_db_ty(
+        &self,
+        storage: &crate::driver::StorageTypes,
+    ) -> crate::Result<crate::schema::db::Type> {
+        use crate::schema::db::Type as DbType;
+
+        let cannot_infer = || {
+            crate::Error::unsupported_feature(format!(
+                "cannot infer a database storage type for {self:?}"
```
MavenRain: This carries user-supplied data, so a failed inference can leak the full payload into the error, and, subsequently, into logs/responses/error-tracking.

Consider `std::mem::discriminant(self)` or a type_name() label, instead of the actual value.

crates/toasty-core/src/stmt/value.rs
```diff
@@ -407,6 +407,101 @@ impl Value {
...
+            // A list stores as the array type of its element, but only when the
+            // elements are uniform at the *app-type* level. Reuse the full
+            // inference path to enforce that: comparing storage types alone is
+            // too permissive, because distinct value types can collapse to one
+            // backend type (e.g. `String` and `Date` both map to TEXT on
+            // SQLite) and a heterogeneous list would slip through. Inferring
+            // through the `TypeUnion` also rejects empty and all-`NULL` lists,
+            // which have no element type.
+            Value::List(_) => {
+                DbType::from_app(&self.infer_ty(), None, storage).map_err(|_| cannot_infer())?
```
MavenRain: Consider preserving the inner error as context: `.map_err(|inner| cannot_infer().context(inner))?`

MavenRain (approved):

crates/toasty-core/src/stmt/value.rs
carllerche: Done, thanks

PR #1036, opened 2026-06-08
fix(cli): include the path in Toasty config load errors

### LINKED ISSUE #1032: Error: No such file or directory (with no Toasty.toml)
### What you did

While running "migration" commands via generated CLI,

1. Even you use Toasty.toml default values, the file is required even in binary. I wish production binary can able to run without Toasty.toml.
2. When developer forgot to copy Toasty.toml in Multi-stage docker builds, while running "migration" commands it gives only "Error: No such file or directory" error. Wish we had a more specific error to identify the issue.

### What you expected

More specific error message to identify missing Toasty.toml

### What actually happened

"Error: No such file or directory"

### Minimal reproducer

1. Build a CLI and run it without Toasty.toml (Common mistake forgetting copy Toasty.toml  on multi-stage build OCI)

### Database driver

Postgressql

### Toasty version

0.7.0

### Additional context

-

gamandeepsingh:
## Summary

`Config::load_from` propagated bare `io::Error`/`toml::de::Error` values
through `?`, so a missing or malformed `Toasty.toml` surfaced only as
`Error: No such file or directory (os error 2)` — with no indication of
*which* path Toasty was looking at. This is especially confusing in
contexts like multi-stage Docker builds, where forgetting to copy
`Toasty.toml` into the final image is an easy mistake to make blind.

This wraps both the read and parse steps with `anyhow::Context`, naming
the path and, for the read failure, hinting at the common Docker cause.

Closes #1032.

## Type of change

- [ ] Other (please explain):

## Checklist

- [x] `cargo test` passes

## Notes for reviewers

`Config::load`/`load_or_default` and the reporter's separate point about
production binaries not requiring `Toasty.toml` at all are left alone —
that's a behavioral question (should `load` fall back to defaults?)
beyond the scope of "the error message doesn't say which file is missing".

### DIFF
--- a/crates/toasty-cli/src/config.rs
+++ b/crates/toasty-cli/src/config.rs
@@ -1,5 +1,5 @@
 use crate::migration::MigrationConfig;
-use anyhow::Result;
+use anyhow::{Context, Result};
 use serde::{Deserialize, Serialize};
 use std::fs;
 use std::path::Path;
@@ -46,8 +46,17 @@ impl Config {
 
     /// Load configuration from a specific path.
     pub fn load_from(path: &Path) -> Result<Self> {
-        let contents = fs::read_to_string(path)?;
-        let config: Config = toml::from_str(&contents)?;
+        let contents = fs::read_to_string(path).with_context(|| {
+            format!(
+                "failed to read Toasty config file at `{}` — check that the file exists \
+                 at this path relative to the working directory (a common cause in Docker \
+                 multi-stage builds is forgetting to copy it into the final image)",
+                path.display()
+            )
+        })?;
+        let config: Config = toml::from_str(&contents).with_context(|| {
+            format!("failed to parse Toasty config file at `{}`", path.display())
+        })?;
         Ok(config)
     }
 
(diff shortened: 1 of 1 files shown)


carllerche (approved):
thanks!

crates/toasty-cli/src/config.rs
```diff
@@ -46,8 +46,17 @@ impl Config {
...
     /// Load configuration from a specific path.
     pub fn load_from(path: &Path) -> Result<Self> {
-        let contents = fs::read_to_string(path)?;
-        let config: Config = toml::from_str(&contents)?;
+        let contents = fs::read_to_string(path).with_context(|| {
+            format!(
+                "failed to read Toasty config file at `{}` — check that the file exists \
+                 at this path relative to the working directory (a common cause in Docker \
+                 multi-stage builds is forgetting to copy it into the final image)",
+                path.display()
```
dumindu, line 54: Feels so AI

gamandeepsingh: fair enough at least the next missing `toasty.toml` should be easier to diagnose

PR #1037, opened 2026-06-09
feat: add set_* replace-variants to Query builder

### LINKED ISSUE #902: feat: add `set_*` replace-variants to `Query` builder
### Problem

PR #899 makes `Query::order_by` additive: chained calls compose into a multi-key sort instead of overwriting. That's the right default, but it leaves no way to overwrite a prior `order_by` without dropping the query and starting over.

The same gap exists for `Query::and`: it composes filters with AND, with no overwrite sibling. `Delete::set_condition` already sets the precedent for an overwrite setter on a filter-shaped field, but `Query` has nothing equivalent.

The rest of the builder surface already follows the convention — `Update` and `Insert` pair additive ops (`insert`, `insert_all`, `remove`) with `set_*` overwrites (`set`, `set_assignments`, `set_scope`), and `CreateMany` is additive-only. `Query` is the outlier.

### Proposed solution

Adopt the convention across builders: bare method names are additive (append, AND, compose); `set_*` prefixed methods replace the field wholesale.

Add two methods on `Query`:

```rust
impl<T> Query<T> {
    pub fn set_order_by(&mut self, order_by: impl Into<stmt::OrderBy>) -> &mut Self;
    pub fn set_filter(&mut self, filter: Expr<bool>) -> &mut Self;
}
```

`set_order_by` writes `self.untyped.order_by = Some(order_by.into())` — the pre-#899 body, kept available under the new name. `set_filter` replaces the filter expression, parallel to `Delete::set_condition`.

Other builders are already aligned and need no change:

| Type | Additive | Replace |
|---|---|---|
| `Update<T>` | `insert`, `remove` | `set`, `set_assignments`, `set_returning_none` |
| `Insert<T>` | `insert`, `insert_all` | `set`, `set_scope` |
| `Delete<T>` | — | `set_condition` |
| `Query<T>` | `and`, `include`, `order_by` (post-#899) | — (this issue) |
| `CreateMany<M>` | `item`, `with_item` | — |

Scalar single-shots (`Query::limit`, `Query::offset`) and mutually-exclusive cursor directives (`Paginate::after`, `Paginate::before`) keep their current bare names — there's no additive semantics to contrast against.

### Alternatives considered

**Add `set_includes` as well.** Skipped for now. Replacing the full include list is a rare operation, and the additive form covers every use case observed so far. Easy to add later if a user asks.

**Rename `latest_by` to keep a `set_latest_by` parallel.** Rejected. `latest_by` is a one-line convenience over `order_by(field.desc())`; users wanting to replace the sort can call `set_order_by` directly.

**Leave `Query::and` alone — only add `set_order_by`.** Viable, but the asymmetry with `Delete::set_condition` is the kind of papercut that surfaces in user code that mixes the two builder shapes. Adding both at once locks in the convention.

### Scope

Small — fits in one PR, no design doc needed. Two new methods on `Query`, additive to the existing surface, no driver impact.

### Additional context

Follow-up to #899 (which makes `order_by` additive) and #803 (the bug that motivated it).

tglane:
## Summary
This adds the methods `set_order_by` and `set_filter` to the `toasty::stmt::Query` type as proposed in #902 to allow for overwriting prior `order_by` or `filter` statements without starting a new query.
Although the changes do touch a public API of the `toasty::stmt::Query` type, #902 clearly states that no design doc is required for this quite small change.

Closes #902 

## Type of change
Small / obvious change — bug fix, docs, internal cleanup, or test

## Checklist

- [x] `cargo test` passes

## Notes for reviewers
This is my first contribution to toasty so if I made some mistakes regarding the contribution conduct, please point me into the right direction for future contributions :)

### DIFF
--- a/crates/toasty-core/src/stmt/query.rs
+++ b/crates/toasty-core/src/stmt/query.rs
@@ -180,6 +180,15 @@ impl Query {
         self.body.as_select_mut_unwrap().add_filter(filter);
     }
 
+    /// Sets the filter for this query's `SELECT` body overwriting any existing one.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the query body is not a `SELECT`.
+    pub fn set_filter(&mut self, filter: impl Into<Filter>) {
+        self.body.as_select_mut_unwrap().filter = filter.into();
+    }
+
     /// Adds an association include path to this query's `SELECT` body.
     pub fn include(&mut self, path: impl Into<Path>) {
         match &mut self.body {
--- a/crates/toasty/src/stmt/query.rs
+++ b/crates/toasty/src/stmt/query.rs
@@ -139,6 +139,27 @@ impl<T> Query<T> {
         self
     }
 
+    /// Sets the filter, combined with AND, for this query overwriting existing ones.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # #[derive(Debug, toasty::Model)]
+    /// # struct User {
+    /// #     #[key]
+    /// #     id: i64,
+    /// #     name: String,
+    /// # }
+    /// use toasty::stmt::{List, Query};
+    ///
+    /// let mut q = Query::<List<User>>::all();
+    /// q.set_filter(User::fields().name().eq("Alice"));
+    /// ```
+    pub fn set_filter(&mut self, filter: Expr<bool>) -> &mut Self {
+        self.untyped.set_filter(filter);
+        self
+    }
+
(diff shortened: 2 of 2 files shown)


carllerche:
Sorry for the delay. Currently on vacation.

carllerche (approved):

tglane:
> Sorry for the delay. Currently on vacation. 

No problem! Enjoy the vacation :)

PR #1038, opened 2026-06-10
feat: introduce #[item_parent] field attribute for item collections

asiefker:
## Summary

Replaces the model-level `#[item_collection(Parent)]` attribute with a field-level `#[item_parent]` on a `Deferred<Parent>` field, and pivots the engine to the symmetric-key model the design has been moving toward: a single hierarchical sort key encodes both row identity and parent membership; the `__model` discriminator column is gone.

```rust
#[derive(toasty::Model)]
#[key(account, sk)]
struct Tenant {
    account: String,
    #[auto] sk: String,
    name: String,
    #[has_many] users: Deferred<Vec<User>>,
}

#[derive(toasty::Model)]
#[key(account, sk)]
struct User {
    account: String,
    #[auto] sk: String,
    name: String,
    #[item_parent] tenant: Deferred<Tenant>,
    #[has_many] todos: Deferred<Vec<Todo>>,
}

#[derive(toasty::Model)]
#[key(account, sk)]
struct Todo {
    account: String,
    #[auto] sk: String,
    title: String,
    #[item_parent] user: Deferred<User>,
}
```

Toasty mints `sk` hierarchically: `Tenant#`, `User#<u-uuid>`, `Todo#<u-uuid>#<t-uuid>`. A partition-scoped `STARTS_WITH` then bounds any subtree, so `acme.users()`, `alice.todos()`, and cascading deletes all reduce to a single DDB query.

The change spans six commits:

1. `feat(macros,core): introduce #[item_parent] field attribute` — parser, schema-build validation that the child's partition + sort fields match the root's by name and type, rejection of the deprecated `#[item_collection(...)]` and named-key forms.
2. `feat(core,engine,macros): symmetric-key engine pivot for item collections` — `FieldTy::ItemParent` and `FieldTy::HasItems` variants, hierarchical sort-key encoding, removal of the `__model` discriminator column, child→parent and parent→children lowering.
3. `test(tests),fix(core,engine): two-tier IC integration tests + engine fixes` — migrates the existing two-tier integration suite and closes three engine gaps the migration surfaced (`Expr::StartsWith` evaluator arms, `HasItems` cascade walker, DDB filter-PK collision).
4. `test(tests,engine): migrate three-tier IC tests with cascade fixes` — migrates the three-tier suite and closes two more engine gaps (simplifier subsumption for nested `StartsWith` predicates; `HasItems` read-scope rewrite accepting `StartsWith`-shaped parent sources from cascade synthesis).
5. `feat(macros): correctness suppressions for item-collection children` — three macro-level guards that turn IC-shape misuses into compile errors.
6. `fix(core,macros),chore: scope #[auto] String + drop dead scaffolding` — restricts `#[auto] String` to item-collection sort keys (it was being granted to any tuple newtype around `String` via the `NewtypeOf` blanket) and drops macro scaffolding that was disconnected post-rebase.

## Type of change

- [x] Implements a previously-discussed direction (item-collection symmetric keys / hierarchical sk; the design has lived in the `item-collection` series of branches).
- [x] Migrates the existing two-tier and three-tier integration tests onto the new surface.
- [x] No public-API breaking change beyond the IC surface itself, which was already private to the IC examples and tests.

## Checklist

- [x] `cargo fmt` and `cargo clippy --all-targets` are clean
- [x] `cargo test -p toasty --lib` (321/321), `cargo test -p toasty-core --tests` (49/49), `cargo test -p tests --test sqlite` (1150/1150)
- [x] `cargo test --features dynamodb item_collection` passes against local DDB (22/22 active, 3 pre-existing deferrals)
- [x] `examples/item-collection` runs end-to-end on SQLite memory and on local DDB
- [x] Trybuild fixtures pin both the deprecated-attribute rejections and the new IC-child correctness suppressions

## Notes for reviewers

- **Hierarchical sk.** A child's sort key is `<Child>#<parent_chain>#<own_uuid>` — strip the parent's model-name prefix, prepend the child's, append a fresh UUID. Read-scope rewriting derives the child prefix from the parent's literal sk via find-first-`#`-and-replace. The encoding is the source of truth; the `apply_lowering_filter_constraint` discriminator (`StartsWith(sk, \"<Model>#\")`) is the safety net for queries that don't go through parent-handle navigation.
- **Discriminator dedup.** Two simplifier rules drop redundant `StartsWith` predicates that the discriminator pass and the IC association lowering each emit independently: `prune_starts_with_subsumed_by_eq` (handles `sk = <full> ∧ StartsWith(sk, \"<Model>#\")`) and `prune_starts_with_subsumed_by_starts_with` (handles two `StartsWith` on the same column, one a prefix of the other). DDB rejects two predicates on the same key, so the dedup is correctness-load-bearing on that backend.
- **Cascade delete.** `acme.delete()` cascades through `HasItems` by synthesizing a `Delete` whose `from` is `Source::Model { via: Some(Association) }` and routing through the existing via-rewrite pipeline. `ItemParent` is a no-op on the cascade walker — the child's PK encoding fixes membership at insert, so there's nothing to null out.
- **Filtered cascades are out of scope.** A cascade through a *filtered* parent set (e.g. `acme.users().filter(name = \"Alice\").delete()`) panics with a clear message at lowering rather than silently over-deleting. The plan doc tracks this as deferred design — three viable shapes (per-parent N+1 deletes, read-then-filter, or reject by design) and no preference yet.
- **Correctness suppressions.** On any model with `#[item_parent]`, the inherent `User::create()` is suppressed (must use `tenant.users().create()...`), and the create-builder no longer exposes `.account(...)` / `.sk(...)` setters. Both turned a confusing runtime panic on sk-encoder mismatch into a compile error. The `create!` macro form falls out for free since it expands to method calls on the create-builder. UI fixtures pin all three.
- **`#[auto] String` scoping.** B4 added `impl Auto for String` so `#[auto] sk: String` would compile on IC models, but the `NewtypeOf` blanket then made any `Tuple(String)` auto-able too. The fix removes the impl and routes IC sort keys through a syntactic detection in the macro instead; schema build rejects any `AutoStrategy::String` that survives the IC promotion pass. A pre-existing UI fixture (`embed_newtype_inner_not_auto.rs`) is restored to passing as a side effect.
- **Deferred follow-ups.** Two ergonomics tasks are intentionally not in this PR; both have workarounds in the existing API:
  - **`get_by_<sk>` / `get_by_local_id` on the relation handle.** Today callers can compose the full sk and call the existing `User::get_by_account_and_sk`. Adding the convenience methods on the relation handle is a separate, smaller PR.
  - **Top-level `Model::get((partition, sk))`.** Same situation — `User::get_by_account_and_sk(&account, &sk)` works today; the tuple-shaped wrapper is convenience.
- **`#[local_id]` is also deferred.** A user-typed local-id segment (`#[local_id] handle: String`) is the natural follow-up after this PR; the IC root sk auto-mint here produces a flat `<Model>#` for roots, which is sufficient for partition-unique roots and the deferred `scoped_isolation` / `scoped_filter_by_id` integration tests un-ignore once `#[local_id]` provides an own-id segment.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ members = [
     # Examples
     "examples/composite-key",
     "examples/hello-toasty",
+    "examples/item-collection",
     "examples/todo-with-cli",
     "examples/user-has-one-profile",
 
@@ -43,6 +44,7 @@ default-members = [
     "crates/toasty-driver-integration-suite-macros",
     "examples/composite-key",
     "examples/hello-toasty",
+    "examples/item-collection",
     "examples/todo-with-cli",
     "examples/user-has-one-profile",
     "tests",
--- a/crates/toasty-core/src/schema/app.rs
+++ b/crates/toasty-core/src/schema/app.rs
@@ -57,7 +57,7 @@ mod pk;
 pub use pk::PrimaryKey;
 
 mod relation;
-pub use relation::{BelongsTo, Cardinality, Has, Via};
+pub use relation::{BelongsTo, Cardinality, Has, HasItems, ItemParent, Via};
 
 mod schema;
 pub use schema::{Resolved, Schema};
--- a/crates/toasty-core/src/schema/app/auto.rs
+++ b/crates/toasty-core/src/schema/app/auto.rs
@@ -21,6 +21,19 @@ pub enum AutoStrategy {
     Uuid(UuidVersion),
     /// Use an auto-incrementing integer sequence (database-assigned).
     Increment,
+    /// Bare UUID v7, written verbatim into a String column.
+    /// Reachable from `impl Auto for String`.
+    String,
+    /// IC root sort key. Emits `<ModelName>#` (literal trailing `#`,
+    /// no own-id segment). Set at schema-build for IC root models.
+    /// The partition column already identifies the row, so no uuid
+    /// is appended. DDB still requires a non-null sort value.
+    ItemCollectionRootSortKey,
+    /// IC descendant sort key. Reads the parent's full sk slot
+    /// (distributed by the HasItems insert-scope walker), swaps the
+    /// model-name prefix with this child's name, and appends
+    /// `#<own-uuid>`. Carries the parent's full local-id chain inline.
+    ItemCollectionChildSortKey,
 }
 
 /// UUID version to use for auto-generated UUID fields.
--- a/crates/toasty-core/src/schema/app/field.rs
+++ b/crates/toasty-core/src/schema/app/field.rs
@@ -2,7 +2,8 @@ mod primitive;
 pub use primitive::{FieldPrimitive, SerializeFormat};
 
 use super::{
-    AutoStrategy, BelongsTo, Constraint, Embedded, Has, Model, ModelId, Schema, VariantId, Via,
+    AutoStrategy, BelongsTo, Constraint, Embedded, Has, HasItems, ItemParent, Model, ModelId,
+    Schema, VariantId, Via,
 };
 use crate::{Result, driver, stmt};
 use std::fmt;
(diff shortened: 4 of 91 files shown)


carllerche:
I'm going to close this. I don't think this is the specific implementation direction needed for the feature. We first should add support for enums as root models, but that requires some other infrastructure first (e.g. #https://github.com/tokio-rs/toasty/pull/1078). Also, I think there are some open questions from the DDB folks about whether or not this is the right API pattern?

PR #1039, opened 2026-06-11
feat: add escape support for like expr

alanskind:
## Summary

Expose the existing `ESCAPE` clause support for `LIKE` / `ILIKE` through the
public query DSL.

`ExprLike` already carries an `escape: Option<char>` field and `toasty-sql`
already serializes it as `LIKE ... ESCAPE ...`, but there was no way to set it
from user code. This adds:

- `Expr::like_with_escape()` / `Expr::ilike_with_escape()` constructors in
  `toasty-core`
- `.like_with_escape(pattern, escape)` / `.ilike_with_escape(pattern, escape)`
  on string-valued field paths in the `toasty` DSL

Both are pass-throughs to the backend's native operator, following the design
philosophy for `.like()` / `.ilike()`: `like_with_escape` works on all SQL
backends with backend-native case sensitivity; `ilike_with_escape` is
PostgreSQL-only and rejected elsewhere with `unsupported_feature`. Toasty does
not escape the pattern automatically — the caller constructs it, and the escape
character makes `%` / `_` matchable as literals.

## Type of change

- [x] Other (please explain): public-API extension mirroring the existing
      `.like()` / `.ilike()` methods; no new driver-layer behavior, the AST
      field and SQL serialization already existed.

## Checklist

- [x] `cargo test` passes
- Touches a public API, but follows the established `.like()` / `.ilike()`
  pass-through pattern, which has no design doc; the `Driver` trait and
  driver-observable behavior are unchanged.

## Notes for reviewers

- Integration tests cover escaping `%`, escaping `_`, and that unescaped
  wildcards in the same pattern still act as wildcards
  (`crates/toasty-driver-integration-suite/src/tests/filter_like.rs`). They are
  gated with `requires(sql)` since DynamoDB has no `LIKE`.
- There is currently no test exercising `ilike_with_escape` against PostgreSQL.

### DIFF
--- a/crates/toasty-core/src/stmt/expr_like.rs
+++ b/crates/toasty-core/src/stmt/expr_like.rs
@@ -50,6 +50,32 @@ impl Expr {
         }
         .into()
     }
+
+    /// Creates a `expr LIKE pattern` expression with escape character.
+    pub fn like_with_escape(expr: impl Into<Self>, pattern: impl Into<Self>, escape: char) -> Self {
+        ExprLike {
+            expr: Box::new(expr.into()),
+            pattern: Box::new(pattern.into()),
+            escape: Some(escape),
+            case_insensitive: false,
+        }
+        .into()
+    }
+
+    /// Creates a `expr ILIKE pattern` expression with escape character.
+    pub fn ilike_with_escape(
+        expr: impl Into<Self>,
+        pattern: impl Into<Self>,
+        escape: char,
+    ) -> Self {
+        ExprLike {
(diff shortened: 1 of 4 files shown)


carllerche (approved):

alanskind:
So, MySQL treats the \ in ESCAPE clause as a escape character

alanskind:
@carllerche Hi, I've fix the mysql case, can you please review it?

carllerche:
Sorry for the delay. I am on vacation.

PR #1047, opened 2026-06-17
docs: add atomic-batches design doc and roadmap entry

asiefker:
## Summary

[Rendered](https://github.com/asiefker/toasty/blob/design/atomic-batches/docs/dev/design/atomic-batches.md)

`toasty::batch()` on DynamoDB executes writes independently — a mid-batch
failure leaves partial state, and cascading deletes can leave dangling
children. This design makes multi-write batches and cascades atomic on
every shipped driver.

A new `Capability::transaction_delivery` names how a driver receives an
atomic group (`Unsupported` / `Streamed` / `WriteSet`). SQL keeps its
existing `BEGIN/COMMIT` streaming; DynamoDB receives a single new
`Operation::TransactWrite` and maps it to `transact_write_items()`. The
guide's "all operations succeed or fail together" promise then holds on
DynamoDB. The doc covers the validation rules (write-on-write rejection,
same-table filter conflicts, duplicate keys), the 100-action cap
(`Error::batch_too_large`), and the DynamoDB driver simplification.

## Type of change

## Checklist

## Notes for reviewers

Design-only, per the propose-then-implement path in `CONTRIBUTING.md`;
implementation lands in a follow-up. The plumbing this targets already
exists in the DynamoDB driver (`transact_write_items()` in six branches)
but is unreachable from `toasty::batch()` today. Worth a close look:
the same-table filter-write rejection (deliberately conservative, with
two refinements deferred to v2) and the two-stage validation split
(structural at plan time, count after the read prefix).

### DIFF
--- a/docs/dev/design/atomic-batches.md
+++ b/docs/dev/design/atomic-batches.md
@@ -1,0 +1,183 @@
+# Atomic Batches for DynamoDB
+
+## Summary
+
+Today, `toasty::batch()` on DynamoDB executes writes independently — a mid-batch failure leaves partial state. This design makes `toasty::batch()` and cascading writes atomic on every backend Toasty ships drivers for. A new driver capability, `Capability::transaction_delivery`, names how a driver receives the writes of an atomic group:
+
+* `Unsupported` - drivers that do not support atomic multi-write operations
+* `Streamed` - the SQL backends, which take the writes one at a time under engine-controlled `BEGIN/COMMIT`
+* `WriteSet` - DynamoDB, which takes them as a single `Operation::TransactWrite` it must commit or cancel together
+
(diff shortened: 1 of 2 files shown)


docs/dev/design/atomic-batches.md
```diff
@@ -0,0 +1,183 @@
+# Atomic Batches for DynamoDB
+
+## Summary
+
+Today, `toasty::batch()` on DynamoDB executes writes independently — a mid-batch failure leaves partial state. This design makes `toasty::batch()` and cascading writes atomic on every backend Toasty ships drivers for. A new driver capability, `Capability::transaction_delivery`, names how a driver receives the writes of an atomic group:
+
+* `Unsupported` - drivers that do not support atomic multi-write operations
+* `Streamed` - the SQL backends, which take the writes one at a time under engine-controlled `BEGIN/COMMIT`
```
carllerche, line 8: I would consider renaming this `Sequential` maybe? Streamed isn't bad, but it means something different to me.

carllerche (approved):

PR #1053, opened 2026-06-25
fix(engine,dynamodb): make multi-key delete and update consistent

asiefker:
## Summary

Multi-key `delete` and `update` on DynamoDB made opposite trade-offs at the transaction boundary:

- **delete** shredded into one `DeleteItem` per key — not atomic.
- **update** sent one `transact_write_items` — atomic, but a filtered update whose subset matched cancelled the whole transaction and updated nothing. It could never update the matching subset the way SQL does.

This shreds both. The engine now fans a multi-key update out into one single-key `UpdateByKey` op per key, mirroring delete, so each key's filter is adjudicated independently and a partial match updates exactly the matching rows. `returning` is preserved by concatenating the per-key rows.

The result is a single easy-to-document contract: multi-key writes are uniformly **non-atomic** on DynamoDB. Atomicity is a deliberate follow-up that will reintroduce it uniformly via the atomic-batches `TransactWrite` work, rather than through these divergent per-operation transact paths.

Also removes the now-unreachable multi-key `transact_write_items` branches from both drivers (the delete branch carried a `todo!()` panic). The single-key and unique-index paths are unchanged.

## Type of change

## Checklist

- [x] `cargo test` passes (SQLite + core; DynamoDB suite run locally against DynamoDB Local)

## Notes for reviewers

Driver-observable behavior changes, but no design doc is added: this aligns delete and update on the existing per-row semantics and removes the lossy filtered-update path — a consistency cleanup, not new public API. The atomic-batches feature that reintroduces atomicity will carry its own design doc.

New test `dynamodb_update_batch_filter::batch_update_filter_partial_match` covers the previously-lossy case (was written test-first and confirmed to fail before the fix). `crud_partitioned::update_by_partition_key` was updated to expect the shredded op sequence.

### DIFF
--- a/crates/toasty/src/engine/exec/update_by_key.rs
+++ b/crates/toasty/src/engine/exec/update_by_key.rs
@@ -48,27 +48,22 @@ impl Exec<'_> {
         // filter is adjudicated independently — matching SQL's per-row
         // semantics, and mirroring how delete fans out. These updates are not
         // atomic.
-        let res = if action.returning.is_some() {
-            let mut rows = vec![];
+        let mut total_count = 0u64;
+        let mut rows = vec![];
 
-            for key in keys {
-                let res = self.exec_update_one(action, key).await?;
-
-                debug_assert!(!res.is_count());
-                rows.extend(res.into_value_stream().collect().await?);
+        for key in keys {
+            match self.exec_update_one(action, key).await? {
+                Rows::Count(n) => total_count += n,
+                other => rows.extend(other.into_value_stream().collect().await?),
             }
+        }
 
+        // The output shape is a property of the action, not the results: with
+        // zero keys there is nothing to match on, yet a `returning` update must
+        // still yield an (empty) stream rather than a count.
+        let res = if action.returning.is_some() {
             Rows::value_stream(ValueStream::from_vec(rows))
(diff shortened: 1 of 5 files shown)


crates/toasty/src/engine/exec/update_by_key.rs
```diff
@@ -44,27 +44,32 @@ impl Exec<'_> {
...
-            let res = self.connection.exec(&self.engine.schema, op.into()).await?;
+            Rows::value_stream(ValueStream::from_vec(rows))
+        } else {
+            let mut total_count = 0u64;
 
-            debug_assert_eq!(!res.values.is_count(), action.returning.is_some());
+            for key in keys {
+                match self.exec_update_one(action, key).await? {
+                    Rows::Count(n) => total_count += n,
+                    _ => panic!("expected Count from UpdateByKey"),
```
MavenRain: I really think that you should return `Err(_)` here instead of panic, as well as exhaustively match the `Rows` enum.

asiefker: That's a good call out. 

The change is building now.

MavenRain (approved):

PR #1055, opened 2026-06-30
docs: design Recency trait with newest/oldest methods

hlsxx:
## Summary

Design proposal for a `Recency` trait and `newest_by` / `oldest_by` methods.

## Type of change

## Notes for reviewers

After some time, I decided to write up a design proposal for the Recency trait and the newest_by / oldest_by sugar methods discussed in [#707](https://github.com/tokio-rs/toasty/pull/707). Working with Toasty regularly, I keep missing a shortcut for ordering by `newest/oldest`, and using Rust's type system to restrict Recency to genuinely orderable columns feels like a strong way to do it — a nice improvement over how similar features are handled in other ORMs (e.g. in PHP or Python ecosystems), where there's typically no compile-time restriction at all.
One naming change from the original PR: I renamed `latest_by `to `newest_by`, so it pairs one-to-one with oldest_by ("newest"/"oldest" share a root and read more symmetrically than "latest"/"oldest").

Happy to discuss further if this is worth pursuing.

### DIFF
--- a/docs/dev/design/recency-ordering.md
+++ b/docs/dev/design/recency-ordering.md
@@ -1,0 +1,165 @@
+# Recency Ordering (`newest_by` / `oldest_by`)
+
+## Summary
+
+Two new query methods, `.newest_by(field)` and `.oldest_by(field)`,
+give a typed shortcut for the common "most/least recent row" query.
+They are sugar over `.order_by(field.desc()/.asc()).first()`, restricted
+at compile time to fields whose type is meaningfully orderable by
+recency via a new marker trait, `stmt::Recency`. The restriction is
+enforced through the type system, not through a runtime check, so
+calling `.newest_by(User::fields().name())` is a compile error rather
+than a query that silently does the wrong thing.
+
+## Motivation
+
+`.order_by(...)` is fully general but says nothing about intent, and
+nothing stops a caller from ordering by a column that isn't a
+meaningful proxy for time:
+
+```rust
+User::all()
+    .order_by(User::fields().name().desc())
+    .first()
+```
+
+This compiles and runs, returning whichever user sorts last
+alphabetically almost certainly not what the caller meant when they
+were trying to find the most recently created user. The bug is easy to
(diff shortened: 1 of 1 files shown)


docs/dev/design/recency-ordering.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+### What fields are accepted
+
+Only fields whose type implements `stmt::Recency` are accepted. Today
+that means timestamp-shaped types — under the `jiff` feature,
+`jiff::Timestamp`, `jiff::Zoned`, `jiff::civil::DateTime`, and
+`jiff::civil::Date`; a matching `chrono` feature gate covers
+`chrono::DateTime<Utc>` and `chrono::NaiveDateTime`. Ordering by an
+arbitrary `String`, `i64`, or `Uuid` field does not compile:
```
joeydewaal, line 85: I think [Carl's note](https://github.com/tokio-rs/toasty/pull/707#issuecomment-4347335294) about the `#[version]` attribute still stands here. An `oldest_by(User::fields().version())` should work IMO. I'd also like type safety with the `Recency` trait but unfortunately I don't think we can.

joeydewaal (commented):
The `newest_by` rename and new `oldest_by` method make sense to me.

docs/dev/design/recency-ordering.md
hlsxx: Yep, I think there could be more edge cases with this proposal, but maybe there is a common way to handle them while keeping the type safety.

One idea I had is that `#[version]` could generate something like `Field<Version<i64>>` where Version<T> implements the Recency trait. For opt-in support on other fields, a `#[recency]` macro could also work. For example it could generate something like `Field<Recency<Timestamp>>`.

joeydewaal: IMO that only complicates things. I'd say stick to `newest_by` and `oldest_by` and remove the `Recency` trait.

joeydewaal:
To clear our backlog I'm cosing this as stale. Consider opening a new issue/pr to continue this discussion if you want.

PR #1065, opened 2026-07-01
feat(engine,sql): support #[version] optimistic concurrency on SQL drivers

### LINKED ISSUE #933: feat: support `#[version]` optimistic concurrency on SQL drivers
## Problem

`#[version]` adds an optimistic-concurrency counter to a model. It works
on DynamoDB but is unavailable on the SQL drivers: every test in
`crates/toasty-driver-integration-suite/src/tests/field_version.rs` is
gated with `requires(not(sql))`. SQL users who want lost-update
protection still hand-roll a version column.

Affected: Toasty users on SQLite, PostgreSQL, and MySQL.

The driver-independent pieces already run on SQL — macro parsing of
`#[version]`, the schema wiring (`version_field`, `versionable`), and
version initialization to `1` on insert
(`crates/toasty/src/engine/lower/insert.rs:154`). The gap is the
conditional write that backs the OCC check.

Removing `requires(not(sql))` and running the suite on SQLite gives 4
passing, 6 failing:

- Pass: `create_initializes_version`, `duplicate_create_fails`,
  `batch_insert_checks_version`, `query_update_multi_key_works`.
- Conditional update — `update_increments_version`, `stale_update_fails`,
  `unique_index_update_increments_version`,
  `unique_index_stale_update_fails`, and `stale_delete_fails` (fails on
  its update step) panic at `crates/toasty/src/engine/mir/project.rs:31`
  (`Option::unwrap()` on `None`). The conditional-update plan produces a
  MIR graph whose `Project` input variable is never assigned.
- Conditional delete — `delete_checks_version` hits `todo!()` at
  `crates/toasty/src/engine/plan/statement.rs:808`. Conditional delete is
  unimplemented for SQL.

## Proposed solution

`docs/dev/design/field-version.md` already describes the intended SQL
behavior: a CTE plan on PostgreSQL, read-modify-write on SQLite and
MySQL, no new capability flag. Closing the gap:

1. **Conditional update on SQL.** The CTE and RMW planners exist
   (`plan_conditional_sql_query_as_cte`,
   `plan_conditional_sql_query_as_rmw` in
   `crates/toasty/src/engine/plan/statement.rs`) and are selected on the
   `cte_with_update` capability, but the generated MIR fails to execute
   for the version case — `Project::to_exec` reads an unset input var.
   Make an instance update with a version condition run end to end.
2. **Conditional delete on SQL.** Route `stmt::Delete.condition` through
   the same two planners. The planner currently `todo!()`s for any
   conditional statement that is not an update (`statement.rs:808`).
3. **Serializer condition handling.** The UPDATE serializer asserts
   `condition.is_none()`
   (`crates/toasty-sql/src/serializer/statement.rs:654`); the DELETE
   serializer ignores `condition` silently. Once the planner rewrites
   conditions into filters before serialization, drop the assertion and
   confirm no condition reaches either serializer.
4. **Query-based version increment.** A query-based update over a
   versioned model should carry `version = version + 1` so any write
   advances the counter (design doc, "Query-based increment"). This is
   unimplemented on every driver today — `query_update_multi_key_works`
   documents that the column is not incremented. Without it, a stale
   instance holder can miss a conflict against a query-based writer.
5. **Un-gate the tests.** Drop `requires(not(sql))` from
   `field_version.rs` so the suite runs on SQLite, PostgreSQL, and MySQL.
   Add coverage for the paths that differ by backend (CTE versus RMW) and
   for concurrent conflicts where the driver locks rows — SQLite reports
   `select_for_update: false`, so its RMW plan reads without a lock.

Ownership enforcement is a prerequisite: a user-supplied value on a
`#[version]` field corrupts the counter. Tracked as the cross-cutting
decision in [#748], which the design doc flags as blocking.

[#748]: https://github.com/tokio-rs/toasty/issues/748

## Alternatives considered

The design doc covers the design-level alternatives (timestamp column,
always-on OCC, builder-level opt-in). For the SQL gap specifically, the
only alternative is keeping `#[version]` DynamoDB-only and documenting it
as such — rejected because the attribute is presented as covering all
four backends and the SQL conditional-write machinery is most of the way
there.

## Scope

Large — cross-cutting, multi-driver.

## Additional context

- Design doc: `docs/dev/design/field-version.md`.
- DynamoDB support is complete; all ten `field_version` tests pass there.
- Related: [#839] (`#[version]` macro rejects a newtype embed of `u64`),
  [#748] (behavior for user-supplied values on Toasty-owned fields).

[#839]: https://github.com/tokio-rs/toasty/issues/839
[#748]: https://github.com/tokio-rs/toasty/issues/748

carllerche:
## Summary

`#[version]` optimistic concurrency worked only on DynamoDB — every test in `field_version.rs` was gated `requires(not(sql))`. This implements the SQL conditional-write path so `#[version]` works on SQLite, PostgreSQL, and MySQL too.

- **Read-modify-write** (SQLite, MySQL) now returns the write's rows, so an instance update reloads its changed fields — including relative updates, via the existing MySQL `UPDATE`→`SELECT` `RETURNING` workaround. The probe projects the condition per matched row (`SELECT <condition> … FOR UPDATE`) instead of an aggregate count, so it can lock exactly the matched rows (`SELECT count(*) … FOR UPDATE` is not universally accepted).
- **CTE plan** (PostgreSQL) compiles a conditional `UPDATE` *or* `DELETE` to a single statement: a `found` probe CTE plus a data-modifying `changed` CTE guarded by `matched = conditioned`. A new `ExprSet::Delete` AST variant lets a `DELETE` be a CTE body, so conditional delete is one round-trip instead of a transaction — matching how conditional `UPDATE` already compiled.
- The SQL serializer now aliases every projected column positionally (`AS column1`, …), fixing a latent bug where CTE/subquery/derived-table column references (`tbl.column1`) resolved against unaliased inner columns. The `UPDATE`/`DELETE` condition assertions become debug asserts — the planner folds the condition into a filter before the statement reaches the serializer. Also fixes a missing space in `INSERT … VALUES (…) RETURNING`.

Closes #933.

## Type of change

- [x] Implements a previously accepted design: [`docs/dev/design/field-version.md`](https://github.com/tokio-rs/toasty/blob/version-for-all/docs/dev/design/field-version.md)

## Checklist

- [x] `cargo test` passes
- [x] Touches driver-observable behavior → the design doc covers it

## Notes for reviewers

- Validated on all four backends: `field_version` 15/15 each; full suites with 0 failures (SQLite 1171, PostgreSQL 1202, MySQL 1180, DynamoDB).
- The serializer change updated several serialization snapshots. The pre-change SQL for CTEs/subqueries referenced `column1` against unaliased inner columns — invalid SQL the serialization-only tests never executed against a database.
- `ExprSet::Delete` is a new statement-AST variant. Its blast radius mirrors the existing `ExprSet::Update`: six exhaustive-match sites in `toasty-core` (visitors, `ExprContext`, `is_const`/`Debug`) plus the SQL `ExprSet` serializer arm.
- A missing-row instance write raises `condition_failed` rather than distinguishing `record_not_found`, matching current DynamoDB behavior (that distinction is implemented on no driver today).

### DIFF
--- a/crates/toasty-driver-integration-suite/src/tests/field_version.rs
+++ b/crates/toasty-driver-integration-suite/src/tests/field_version.rs
@@ -1,7 +1,7 @@
 use crate::prelude::*;
 
 /// A newly created record starts with version == 1.
-#[driver_test(requires(not(sql)), scenario(crate::scenarios::versioned_item))]
+#[driver_test(scenario(crate::scenarios::versioned_item))]
 pub async fn create_initializes_version(test: &mut Test) -> Result<()> {
     let mut db = setup(test).await;
 
@@ -14,7 +14,7 @@ pub async fn create_initializes_version(test: &mut Test) -> Result<()> {
 }
 
 /// Updating a record increments the version.
-#[driver_test(requires(not(sql)), scenario(crate::scenarios::versioned_item))]
+#[driver_test(scenario(crate::scenarios::versioned_item))]
 pub async fn update_increments_version(test: &mut Test) -> Result<()> {
     let mut db = setup(test).await;
 
@@ -38,7 +38,7 @@ pub async fn update_increments_version(test: &mut Test) -> Result<()> {
 /// returning projection. The two must not collide: the engine asks the driver
 /// for exactly the relative column, so the version value never lands in the
 /// returned row and shifts `value` out of its slot.
-#[driver_test(requires(not(sql)), scenario(crate::scenarios::versioned_counter))]
+#[driver_test(scenario(crate::scenarios::versioned_counter))]
 pub async fn relative_update_increments_value_and_version(test: &mut Test) -> Result<()> {
     let mut db = setup(test).await;
 
@@ -64,7 +64,7 @@ pub async fn relative_update_increments_value_and_version(test: &mut Test) -> Re
 
 /// Two updates from the same stale snapshot — the second should fail with a
 /// condition-check error because the DB version has already moved to 2.
-#[driver_test(requires(not(sql)), scenario(crate::scenarios::versioned_item))]
+#[driver_test(scenario(crate::scenarios::versioned_item))]
 pub async fn stale_update_fails(test: &mut Test) -> Result<()> {
     let mut db = setup(test).await;
 
@@ -93,7 +93,7 @@ pub async fn stale_update_fails(test: &mut Test) -> Result<()> {
 
 /// Creating the same primary key twice should fail because of the
 /// attribute_not_exists condition on the version column.
-#[driver_test(requires(not(sql)), scenario(crate::scenarios::versioned_item))]
+#[driver_test(scenario(crate::scenarios::versioned_item))]
 pub async fn duplicate_create_fails(test: &mut Test) -> Result<()> {
     let mut db = setup(test).await;
 
@@ -118,7 +118,7 @@ pub async fn duplicate_create_fails(test: &mut Test) -> Result<()> {
 
 /// Batch-creating multiple versioned items should initialize all versions to 1,
 /// and a duplicate within the batch should fail.
-#[driver_test(requires(not(sql)), scenario(crate::scenarios::versioned_item))]
+#[driver_test(scenario(crate::scenarios::versioned_item))]
 pub async fn batch_insert_checks_version(test: &mut Test) -> Result<()> {
     let mut db = setup(test).await;
 
@@ -156,7 +156,7 @@ pub async fn batch_insert_checks_version(test: &mut Test) -> Result<()> {
 /// Unlike an instance update there is no per-row OCC guard — a query-based
 /// update is atomic at the database level — but the version still advances so
 /// concurrent stale writers are detected.
-#[driver_test(requires(not(sql)), scenario(crate::scenarios::versioned_item))]
+#[driver_test(scenario(crate::scenarios::versioned_item))]
 pub async fn query_update_increments_version(test: &mut Test) -> Result<()> {
     let mut db = setup(test).await;
 
@@ -181,7 +181,7 @@ pub async fn query_update_increments_version(test: &mut Test) -> Result<()> {
 /// A query-based update bumps the version, so a concurrent instance update from
 /// a snapshot taken *before* the query update fails its OCC check rather than
 /// silently clobbering the query update's write.
-#[driver_test(requires(not(sql)), scenario(crate::scenarios::versioned_item))]
+#[driver_test(scenario(crate::scenarios::versioned_item))]
 pub async fn query_update_invalidates_stale_instance(test: &mut Test) -> Result<()> {
     let mut db = setup(test).await;
 
@@ -222,7 +222,7 @@ pub async fn query_update_invalidates_stale_instance(test: &mut Test) -> Result<
 /// The increment is applied atomically to every matched row, so each row's
 /// version advances independently. Verifies the multi-key transact path applies
 /// all assignments and bumps each version.
-#[driver_test(requires(not(sql)), scenario(crate::scenarios::tagged_item))]
+#[driver_test(scenario(crate::scenarios::tagged_item))]
 pub async fn query_update_multi_key_works(test: &mut Test) -> Result<()> {
     let mut db = setup(test).await;
 
@@ -255,10 +255,7 @@ pub async fn query_update_multi_key_works(test: &mut Test) -> Result<()> {
 /// Query-based update through the unique-index path (path 3) increments the
 /// version. The version is a non-unique column, so it rides along in the main
 /// update expression alongside the unique-index surgery.
-#[driver_test(
-    requires(not(sql)),
-    scenario(crate::scenarios::versioned_user_unique_email)
-)]
+#[driver_test(scenario(crate::scenarios::versioned_user_unique_email))]
 pub async fn query_update_unique_index_increments_version(test: &mut Test) -> Result<()> {
     let mut db = setup(test).await;
 
@@ -284,10 +281,7 @@ pub async fn query_update_unique_index_increments_version(test: &mut Test) -> Re
 
 /// Updating a record through the unique-index path (path 3) increments the
 /// version when the unique column changes.
-#[driver_test(
-    requires(not(sql)),
-    scenario(crate::scenarios::versioned_user_unique_email)
-)]
+#[driver_test(scenario(crate::scenarios::versioned_user_unique_email))]
 pub async fn unique_index_update_increments_version(test: &mut Test) -> Result<()> {
     let mut db = setup(test).await;
 
@@ -315,10 +309,7 @@ pub async fn unique_index_update_increments_version(test: &mut Test) -> Result<(
 
 /// Stale update on a model with a unique index: the second update from a stale
 /// snapshot should fail.
-#[driver_test(
-    requires(not(sql)),
-    scenario(crate::scenarios::versioned_user_unique_email)
-)]
+#[driver_test(scenario(crate::scenarios::versioned_user_unique_email))]
 pub async fn unique_index_stale_update_fails(test: &mut Test) -> Result<()> {
     let mut db = setup(test).await;
 
@@ -359,7 +350,7 @@ pub async fn unique_index_stale_update_fails(test: &mut Test) -> Result<()> {
 /// returning column list; the engine-injected version bump is kept *out* of
 /// that list. The driver returns exactly the requested columns, so the version
 /// value never lands in the returned row and shifts `value` out of its slot.
-#[driver_test(requires(not(sql)), scenario(crate::scenarios::versioned_counter))]
+#[driver_test(scenario(crate::scenarios::versioned_counter))]
 pub async fn query_relative_update_increments_value_and_version(test: &mut Test) -> Result<()> {
     let mut db = setup(test).await;
 
@@ -379,7 +370,7 @@ pub async fn query_relative_update_increments_value_and_version(test: &mut Test)
 }
 
 /// Deleting a record checks the version — a fresh handle succeeds.
-#[driver_test(requires(not(sql)), scenario(crate::scenarios::versioned_item))]
+#[driver_test(scenario(crate::scenarios::versioned_item))]
 pub async fn delete_checks_version(test: &mut Test) -> Result<()> {
     let mut db = setup(test).await;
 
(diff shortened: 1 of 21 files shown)


crates/toasty/src/engine/plan/statement.rs
```diff
@@ -1027,209 +1036,213 @@ impl<'a, 'b> PlanStatement<'a, 'b> {
...
         };
 
-        let Some(filter) = stmt.filter.expr else {
-            panic!("conditional update without filter");
-        };
+        // `changed`: the write. It applies only when the probe's two counts
+        // agree, expressed as a scalar subquery over `found` ANDed onto the
+        // filter. `found` sits two scopes out from here (subquery → CTE body →
+        // WITH).
+        let counts_agree = stmt::Expr::stmt(stmt::Select {
```
carllerche, line 1083: **[correctness — confirmed]** The CTE strategy's OCC check races under READ COMMITTED. The `found` probe evaluates the version condition against the statement snapshot without locking, and when the write blocks on a concurrent committed update, PostgreSQL's re-check (EvalPlanQual) only re-evaluates the PK filter plus the already-materialized probe subquery — so a stale writer whose probe passed silently overwrites the newer row, and both callers see success. This is the lost update `#[version]` exists to prevent (SQLite/MySQL are safe via the RMW path's `SELECT ... FOR UPDATE` probe).

Suggested fix: make the probe a per-row `SELECT <condition> FROM t WHERE <filter> FOR UPDATE` CTE (aggregate-free, so the locking clause is legal) and derive the matched/conditioned counts in a second CTE over it. The lock forces the condition to be evaluated against the latest committed row version, closing the race.

crates/toasty/src/engine/plan/statement.rs
```diff
@@ -1027,209 +1036,213 @@ impl<'a, 'b> PlanStatement<'a, 'b> {
...
-                .build(),
-        });
-
-        let returning_len = match &stmt.returning {
-            Some(stmt::Returning::Project(expr)) => {
-                let stmt::Expr::Record(expr_record) = expr else {
-                    panic!("returning must be a record");
-                };
+        // `changed`: the write, wrapped as a data-modifying CTE body.
+        let changed_body: stmt::ExprSet = match write {
```
carllerche, line 1124: **[correctness — confirmed]** Bind params inside this data-modifying CTE body escape type refinement: `extract_params::refine_query` only recurses into `Select`/`Values` CTE bodies, so params in an `ExprSet::Update`/`ExprSet::Delete` body whose type can't be inferred from the value alone (`f64`, `f32`, …) stay `Ty::Unknown` and `finalize_ty` panics.

Repro (PostgreSQL): a `#[version]` model with an `f64` field — `item.update().score(2.5).exec(...)` — panics with `extract_params left F64(2.5) with unresolved type` (extract_params.rs:77). The same update passes on the RMW backends because there the write stays a top-level `Statement::Update` that `refine_stmt` handles.

Suggested fix: add `ExprSet::Update` / `ExprSet::Delete` arms to `refine_query`.

crates/toasty/src/engine/exec/exec_statement.rs
```diff
@@ -145,58 +166,91 @@ impl Exec<'_> {
...
+                } else if let Some(mysql_update) = mysql_update_returning {
+                    res = self
+                        .run_mysql_update_returning_select(mysql_update, action.output.ty.clone())
+                        .await?;
+                }
+            }
+            ConditionalOutput::Count => {
+                let rows = collect_conditional_probe(res.values).await?;
+                let (matched, conditioned) = conditional_probe_counts(&rows[0])?;
+                if matched != conditioned {
```
carllerche: **[correctness — confirmed]** When the probe matches zero rows, `0 == 0` passes and the conditional write reports success. The design doc (field-version.md, "Row missing vs. version mismatch") calls for `Error::record_not_found` when the row is gone, and DynamoDB errors on this case — so the same stale-instance update after a concurrent delete returns Ok on SQL and Err on DynamoDB, contradicting the guide's "the behavior is the same everywhere". The RMW path has the identical hole (`exec/rmw.rs:149`).

Suggested fix: `matched == 0` → `Error::record_not_found` in both conditional paths.

crates/toasty/src/engine/exec/exec_statement.rs
```diff
@@ -145,58 +166,91 @@ impl Exec<'_> {
...
+                    return Err(toasty_core::Error::condition_failed(
+                        "write condition did not match",
+                    ));
+                }
+
+                // When nothing matched, the LEFT JOIN yields a single
+                // NULL-padded row; drop it. Otherwise every row is a real
+                // changed row — strip the two leading probe columns.
+                let changed = if matched == 0 {
```
carllerche: **[correctness — race]** "Nothing matched" is judged by the probe count, not by whether `changed` actually produced rows. If the probe matched but the CTE UPDATE ended up changing nothing (row concurrently deleted between the snapshot and the write), the single NULL-padded LEFT JOIN row is treated as a real changed row and a record of all-NULL fields is handed to the instance-update reload.

This falls out of the same root cause as the unlocked probe: with a `FOR UPDATE` probe, matched rows can't disappear before the write, and this case becomes unreachable.

crates/toasty/src/engine/exec/rmw.rs
```diff
@@ -142,22 +175,44 @@ impl Exec<'_> {
...
+        }
+
         let Rows::Count(actual) = res.values else {
             return Err(toasty_core::Error::invalid_result(
                 "expected Count, got Stream",
             ));
         };
 
-        assert_eq!(actual, count as u64);
+        assert_eq!(actual, count);
```
carllerche: **[robustness]** `assert_eq!(actual, count)` panics mid-transaction when the write's affected-row count diverges from the probe count. On SQLite (`select_for_update` is false) the probe takes no row locks, so a concurrent writer can change the matched set between the probe and the write — that should surface as a retryable `Err` (the caller already handles rollback), not a process abort. Every other unexpected-result case in this function returns an error.

crates/toasty/src/engine/mir/read_modify_write.rs
```diff
@@ -40,15 +40,29 @@ impl ReadModifyWrite {
...
-        // A hack since rmw doesn't support output yet
-        let var = var_table.register_var(stmt::Type::list(stmt::Type::Unit));
+        let var = var_table.register_var(self.ty.clone());
+        node.var.set(Some(var));
+
+        // When the write carries a `RETURNING`, the node's type is
+        // `List<Record>`; the record field types tell the driver how to decode
+        // the returned rows. Without a `RETURNING` the type is `Unit` and the
+        // write reports only a row count.
+        let output_ty = match &self.ty {
```
carllerche: **[cleanup]** This `List<Record>` → field types / `Unit` → `None` match duplicates `mir/exec_statement.rs:59-70` introduced in the same diff; the two `todo!`-guarded matches will drift. A shared helper (e.g. `fn row_field_types(ty: &stmt::Type) -> Option<Vec<stmt::Type>>`) removes the duplication.

crates/toasty/src/engine/exec/exec_statement.rs
```diff
@@ -145,58 +166,91 @@ impl Exec<'_> {
...
+            ConditionalOutput::None => {
+                if let Some(mysql_info) = mysql_insert_returning {
+                    res.values = mysql_info.reconstruct_returning(res.values).await?;
+                } else if let Some(mysql_update) = mysql_update_returning {
+                    res = self
+                        .run_mysql_update_returning_select(mysql_update, action.output.ty.clone())
+                        .await?;
+                }
+            }
+            ConditionalOutput::Count => {
```
carllerche: **[cleanup]** The `Count` and `Returning` arms duplicate the probe collection, count extraction, and mismatch check verbatim (lines 214-220 vs 224-230, including the error string). Hoisting the shared prefix above the match — only the `None` arm doesn't need it — keeps the probe protocol and error message in one place.

crates/toasty/src/engine/plan/statement.rs
```diff
@@ -1845,6 +1858,31 @@ impl<'a, 'b> PlanStatement<'a, 'b> {
...
+/// (both at zero nesting).
+fn cte_column(table: usize, column: usize) -> stmt::Expr {
+    stmt::Expr::column(stmt::ExprColumn {
+        nesting: 0,
+        table,
+        column,
+    })
+}
+
+/// The probe projection shared by both SQL conditional-write strategies:
```
carllerche: **[docs]** The doc comment says this projection is "shared by both SQL conditional-write strategies", but this same diff rewrote `plan_conditional_sql_query_as_rmw` to project the condition per matched row (for `FOR UPDATE` compatibility) — the CTE planner is now the sole caller. A maintainer following this doc could wrongly "re-share" it into the RMW probe.

crates/toasty-sql/src/serializer/statement.rs
```diff
@@ -265,6 +265,15 @@ impl ToSql for &stmt::Delete {
     fn to_sql(self, f: &mut super::Formatter<'_>) {
         assert!(self.returning.is_none());
 
+        // Conditions never reach the serializer: a conditional DELETE is
```
carllerche: **[docs]** This comment points at `plan_conditional_sql_query_as_rmw`, but the reason `ExprSet::Delete` exists — and the only path that serializes a DELETE inside a query — is the CTE planner. The sibling UPDATE comment (line ~655) says `plan_conditional_sql_query_as_*`; the same wording fits here.

crates/toasty-driver-integration-suite/src/tests/field_version.rs
```diff
@@ -93,7 +93,7 @@ pub async fn stale_update_fails(test: &mut Test) -> Result<()> {
 
 /// Creating the same primary key twice should fail because of the
 /// attribute_not_exists condition on the version column.
-#[driver_test(requires(not(sql)), scenario(crate::scenarios::versioned_item))]
+#[driver_test(scenario(crate::scenarios::versioned_item))]
```
carllerche, line 97: **[docs]** Now that this test runs on all backends, the doc comment's "attribute_not_exists condition on the version column" only explains DynamoDB — on SQL backends the duplicate create fails via the primary-key constraint. Worth generalizing so someone debugging a SQL failure isn't pointed at the wrong mechanism. (Same issue on `query_update_multi_key_works`, whose doc describes the DDB transact path.)

crates/toasty-driver-integration-suite/src/tests/field_version.rs
```diff
@@ -222,7 +222,7 @@ pub async fn query_update_invalidates_stale_instance(test: &mut Test) -> Result<
 /// The increment is applied atomically to every matched row, so each row's
 /// version advances independently. Verifies the multi-key transact path applies
 /// all assignments and bumps each version.
-#[driver_test(requires(not(sql)), scenario(crate::scenarios::tagged_item))]
+#[driver_test(scenario(crate::scenarios::tagged_item))]
```
carllerche, line 227: **[docs]** "update_by_key path 2 (no unique index, N keys via transact_write_items on DDB)" is DynamoDB-specific; on SQL backends this is a plain multi-row UPDATE. Worth a driver-neutral phrasing now that the test is un-gated from `requires(not(sql))`.

PR #1072, opened 2026-07-04
feat: add support for Turso Sync in toasty-driver-turso

CCXLV:
## Summary

Adds support for Turso Sync with new feature `sync` on `toasty-driver-turso`

Usage:

```rust
let driver = toasty_driver_turso::Turso::file("database.db")
        .with_remote_url("http://127.0.0.1:8080");
   
let db = toasty::Db::builder()
        .models(toasty::models!(crate::*))
        .build(driver.clone())
        .await?;

driver.push().await?;
``` 

Features needed to be implemented:

- [x] with_remote_url
- [x] with_auth_token
- [x] with_auth_token_fn
- [x] with_client_name
- [x] with_long_poll_timeout
- [x] with_remote_encryption
- [x] with_remote_encryption_key
- [x] bootstrap_if_empty
- [x] experimental_with_partial_sync_opts
- [x] push
- [x] pull
- [x] checkpoint
- [x] stats

## Type of change

- [ ] Other (please explain):

## Checklist

- [x] `cargo test` passes


.github/actions/setup-turso-sync/action.yml
```diff
@@ -0,0 +1,30 @@
...
+    value: ${{ steps.start.outputs.url }}
+
+runs:
+  using: composite
+  steps:
+    - name: Install tursodb
+      shell: bash
+      run: |
+        curl --proto '=https' --tlsv1.2 -LsSf \
+          https://github.com/tursodatabase/turso/releases/download/v0.6.1/turso_cli-installer.sh | sh
```
glommer: I am not sure pinning the version here is good. Should we at least pin to 0.7.0 (released today) ?

crates/toasty-driver-turso/src/lib.rs
```diff
@@ -96,24 +109,48 @@ enum TursoPath {
...
 /// [`turso::Builder`] at connection time.
+#[cfg(not(feature = "sync"))]
 #[derive(Debug, Default, Clone)]
-struct BuilderOptions {
+struct LocalBuilderOptions {
     encryption: Option<EncryptionOpts>,
     attach: bool,
     custom_types: bool,
     generated_columns: bool,
-    index_method: bool,
```
glommer, line 109: what is special about index_method that required it to be treated differently here ?

crates/toasty-driver-turso/src/lib.rs
```diff
@@ -161,21 +276,18 @@ impl BuilderOptions {
 /// ```
 /// use toasty_driver_turso::Turso;
 ///
-/// // In-memory database
```
glommer, line 164: why remove this example ?

crates/toasty-driver-turso/src/lib.rs
```diff
@@ -255,71 +367,245 @@ impl Turso {
...
-        self.options.without_rowid = on;
+        self.options.local_options.without_rowid = on;
+        self
+    }
+
+    /// Set the remote base URL for sync HTTP requests. Mirrors
+    /// `turso::sync::Builder::with_remote_url`.
+    ///
+    /// Accepts `https://`, `http://`, and `libsql://` URLs (`libsql://` is
+    /// translated to `https://`). If omitted on a file-backed database that
```
glommer, line 479: should accept turso:// as well, we'll start standardizing on it more.

glommer (changes requested):
I'd like to at least understand the reason for why the index method boolean got its own treatment. It looks straightforward otherwise.

.github/actions/setup-turso-sync/action.yml
CCXLV: I did that because of the version that was used in the root `Cargo.toml` and didn't think it would be good to run the tests on latest version.

```toml
turso = "0.6"
```

crates/toasty-driver-turso/src/lib.rs
CCXLV: Because it was used in both sync and non sync features and I decided to do that just to avoid code duplication, since I had created a different struct for non sync feature options.

crates/toasty-driver-turso/src/lib.rs
CCXLV: I actually messed that one up, will fix it

CCXLV:
Should I add more detailed tests?

CCXLV:
Just updated the docs

glommer (approved):

.github/actions/setup-turso-sync/action.yml
carllerche: Is the download version and crate version in sync? If yes, we can just reuse that. Otherwise, we will need an automated way to update it.

carllerche:
@codex review this

.github/actions/setup-turso-sync/action.yml
CCXLV: No, it would need a manual update in case of version change.

CCXLV: I'll see what I can do to make it download the crate's version automatically.

crates/toasty-driver-turso/src/lib.rs
```diff
@@ -255,71 +399,245 @@ impl Turso {
...
     pub fn experimental_without_rowid(mut self, on: bool) -> Self {
-        self.options.without_rowid = on;
+        self.options.local_options.without_rowid = on;
+        self
+    }
+
+    /// Set the remote base URL for sync HTTP requests. Mirrors
+    /// `turso::sync::Builder::with_remote_url`.
+    ///
+    /// Accepts `https://`, `http://`, `turso://`, and `libsql://` URLs (`libsql://` is
```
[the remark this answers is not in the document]
glommer: turso is a valid scheme and will be our preferred way going forward, fwiw

.github/actions/setup-turso-sync/action.yml
CCXLV: I synced the version to the crate's version in the recent commit.

crates/toasty-driver-turso/src/lib.rs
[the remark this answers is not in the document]
CCXLV: I just checked it and actually current implementation doesn't allow `turso://`

https://github.com/tursodatabase/turso/blob/fbc64afb2c55aa8b45faa8b13010491518c1f804/bindings/rust/src/sync.rs#L595-L610

```rust
// Normalize remote base URL, mapping libsql:// to https:// and validating allowed schemes.
fn normalize_base_url(input: &str) -> std::result::Result<String, String> {
    let s = input.trim();
    let s = if let Some(rest) = s.strip_prefix("libsql://") {
        format!("https://{rest}")
    } else {
        s.to_string()
    };
    // Accept http or https only
    if !(s.starts_with("https://") || s.starts_with("http://")) {
        return Err(format!("unsupported remote URL scheme: {input}"));
    }
    // Ensure no trailing slash to make join predictable.
    let base = s.trim_end_matches('/').to_string();
    Ok(base)
}
```

joeydewaal (approved):

PR #1085, opened 2026-07-13
fix(macros): normalize raw identifiers in collision checks

### LINKED ISSUE #989: macros: raw-identifier field names bypass projection-method collision check
### Summary

Field names written as raw identifiers (`r#count`, `r#insert`, …) bypass the
collision check that #987 added for the generated `Query`/`Many`/`One`/`OptionOne`
projection methods. The macro emits the projection method anyway, which then
clashes with the existing wrapper method and produces a duplicate-inherent-method
compile error instead of the intended skip + deprecation warning.

Root cause: both check sites compare `field_ident.to_string()` against
`QUERY_RESERVED_METHODS` / `MANY_RESERVED_METHODS` / `ONE_RESERVED_METHODS`.
`syn::Ident::to_string()` retains the leading `r#` for raw identifiers, so
`r#count` is compared as `"r#count"` and never matches `"count"`. Rust then
treats the emitted `fn r#count` and the existing `fn count` as the same method.

Affected sites:

- `crates/toasty-macros/src/model/expand.rs:60` — `projection_method_collides`
  (used by the `Query` projection-method expansion).
- `crates/toasty-macros/src/model/expand/relation.rs:336` —
  `reserved.contains(&field_ident.to_string().as_str())` in
  `expand_wrapper_projection_methods` for `Many` / `One` / `OptionOne`.

Both should strip a leading `r#` before the comparison.

### Reproducer

```rust
struct Item {
    #[key]
    #[auto]
    id: i64,
    r#count: i64,
}
```

Expected: macro skips the projection method, emits the `#[deprecated]` marker;
user gets a compile-time warning pointing at `.select(Item::fields().count())`.

Actual: macro emits `fn r#count(self) -> …`, colliding with the wrapper's
existing `count` method (or, with `Query::count_rows` since the rename, the
collision is with whichever wrapper still defines `count` after future
additions). Compile fails with a duplicate-inherent-method error.

### Database driver

none — bug is in `toasty-macros`.

### Toasty version

main (current `worktree-floofy-plotting-pie` tip, ahead of `198b7d6d`).

### Additional context

Follow-up from review of #987
(https://github.com/tokio-rs/toasty/pull/987#discussion_r3314430943). The PR
introduced the collision-check machinery; this bug exists in that code path
from the start.

Pritam-deb:
## Summary

Closes #989.

Raw identifier fields like `r#count` stringify as `r#count`, but Rust treats the generated method name as `count`. This caused generated projection/relation shortcut collision checks to miss reserved method names and emit duplicate inherent methods.

This PR normalizes identifiers before reserved-method checks and skips colliding generated shortcuts on model, query, and field-path impls. It also adds a trybuild pass fixture for raw identifiers that normalize to reserved method names.

## Type of change

## Checklist

- [x] `cargo test` passes

## Notes for reviewers

Verified with:

- `cargo test -p tests ui_pass`
- `cargo test -p tests ui`
- `cargo test -p toasty-driver-integration-suite raw_identifier_fields --no-run`
- `cargo fmt`

### DIFF
--- a/crates/toasty-macros/src/model/expand/fields.rs
+++ b/crates/toasty-macros/src/model/expand/fields.rs
@@ -4,6 +4,11 @@ use crate::model::schema::ModelKind;
 use proc_macro2::TokenStream;
 use quote::{quote, quote_spanned};
 
+const FIELD_STRUCT_RESERVED_METHODS: &[&str] =
+    &["from_path", "path", "eq", "in_query", "into_root", "create"];
+
+const FIELD_LIST_STRUCT_RESERVED_METHODS: &[&str] = &["from_path", "path", "any", "all", "create"];
+
 impl Expand<'_> {
     pub(super) fn expand_field_struct(&self) -> TokenStream {
         let toasty = &self.toasty;
@@ -32,6 +37,9 @@ impl Expand<'_> {
             .fields
             .iter()
             .enumerate()
+            .filter(|(_, field)| {
+                !util::ident_is_reserved(&field.name.ident, FIELD_STRUCT_RESERVED_METHODS)
+            })
             .map(move |(offset, field)| {
                 let field_ident = &field.name.ident;
                 let field_offset = util::int(offset);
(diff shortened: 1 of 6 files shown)


carllerche (approved):
thanks!

PR #1088, opened 2026-07-13
docs: document embedded enum rename_all and string discriminants

carllerche:
## Summary

The guide and the `Embed` rustdoc still described embedded enums as
requiring an integer `#[column(variant = N)]` on every variant. Since
#1082 and #1083, variants default to a `snake_case` string label,
`#[column(rename_all = "...")]` selects another naming rule, and
`#[column(variant = "...")]` overrides a single label. Integer
discriminants remain supported but are now the opt-in path.

Both documents now lead with string labels, list the eight supported
`rename_all` rules, and cover the constraints: string and integer
discriminants cannot be mixed, `rename_all` applies only to string
labels, and integer discriminants must be given on every variant.
The guide also drops the hand-written `CREATE TABLE` snippets, which
claimed an `INTEGER` discriminant column and no longer matched what
Toasty emits.

## Type of change

## Checklist

- [x] `cargo test` passes

## Notes for reviewers

Docs only — no code changes. Worth checking that the `rename_all` rule
table matches the conversions the macro actually performs.

### DIFF
--- a/crates/toasty-macros/src/lib.rs
+++ b/crates/toasty-macros/src/lib.rs
@@ -789,34 +789,31 @@ pub fn derive_model(input: TokenStream) -> TokenStream {
 /// # Enums
 ///
 /// An embedded enum stores a discriminant value identifying the active
-/// variant. Each variant must have a `#[column(variant = N)]` attribute
-/// assigning a stable integer discriminant.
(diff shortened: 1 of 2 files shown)


carllerche:
@codex review this again

PR #1089, opened 2026-07-13
feat: filter associations in include

joeydewaal:
## Summary
With this pr it's possible to filter included associations like:
```rust
let users: Vec<User> = User::all()
    .include(
        User::fields()
            .todos()
            .filter(Todo::fields().title().eq("keep")),
    )
    .exec(&mut db)
    .await?;
```
Note this diverges from the original design doc, see "Notes for reviewers".

## Type of change

      (https://github.com/tokio-rs/toasty/blob/main/docs/dev/design/include-filters.md):
- [ ] Other (please explain):

## Checklist

- [x] `cargo test` passes

## Notes for reviewers
Since associations should be treated as a set, in the original plan there is a section about [distinguishing loaded from partial](https://github.com/tokio-rs/toasty/blob/main/docs/dev/design/include-filters.md#distinguishing-loaded-from-partial).
A while back support was added for eager associations. We could add a new `Partial` state to the `Deferred` enum like mentioned in the plan but to me this seems inconsistent. That would mean only deferred associations would be treated as a set.
There's no way to differentiate between `Loaded` and `Partial` in eager associations since it's just a `Vec`.

I opted to remove the `Partial` state entirely. This makes filtering included association consistent between eager and deferred associations but this also means that associations are not treated as a set.

Edit:
One think I'm unhappy about is; filtering on 1-1 relations should not be allowed unless the relation is optional. Currently this is a runtime error but ideally this should be a compile time error. This should be possible but I've left this for a follow up pr.

Via support is also for a followup pr.


carllerche:
Ping me when this I ready for review.

joeydewaal:
@carllerche this is ready for a review.

crates/toasty-driver-integration-suite/src/tests/relation_preload_filter.rs
```diff
@@ -0,0 +1,579 @@
...
+    assert_eq!(1, users.len());
+    let alice = &users[0];
+    assert_eq!("alice", alice.name);
+    let titles: Vec<&str> = alice.todos.get().iter().map(|t| t.title.as_str()).collect();
+    assert_eq!(titles, vec!["keep"]);
+
+    Ok(())
+}
+
+/// Multiple `.include(...)` calls on the same relation with different
```
carllerche: The filters look the same in this test. That said, if you are *stacking* `.include(...)` for the same field, the predicates should probably be combined with `OR`, not `AND`.

crates/toasty-driver-integration-suite/src/tests/relation_preload_filter.rs
```diff
@@ -0,0 +1,579 @@
...
+
+    Ok(())
+}
+
+/// A bare `.include(todos())` (load all) combined with a filtered
+/// `.include(todos().filter(...))` on the same relation loads the filtered
+/// set: the predicate wins. An unfiltered include contributes no predicate,
+/// so it does not broaden the result back to all.
+#[driver_test(id(ID), scenario(crate::scenarios::has_many_belongs_to))]
+pub async fn preload_has_many_bare_and_filtered_uses_filter(test: &mut Test) -> Result<()> {
```
carllerche: Same here, IMO combining `.include(..)` should be `OR`

crates/toasty-driver-integration-suite/src/tests/relation_preload_filter.rs
```diff
@@ -0,0 +1,579 @@
...
+    assert_eq!(bodies, vec!["keep"]);
+
+    Ok(())
+}
+
+/// An eager (non-deferred) relation is an implicit bare include, so a
+/// filtered `.include(...)` of it restricts what loads — the implicit
+/// include contributes no predicate and does not broaden the result.
+#[driver_test]
+pub async fn preload_eager_has_many_with_filter(test: &mut Test) -> Result<()> {
```
carllerche: I disagree with this. An eager relation states that everything should be loaded. This goes back to AND vs. OR, and I think OR makes more sense.

crates/toasty-core/src/stmt/include.rs
```diff
@@ -0,0 +1,47 @@
...
+/// than an `Expr` so the predicate's scope travels with it (the same
+/// mechanism `Path::any` relies on); lowering extracts the `WHERE` clause
+/// and ignores everything else.
+#[derive(Debug, Clone, PartialEq)]
+pub struct Include {
+    /// The relation path to preload.
+    pub path: Path,
+
+    /// Optional filter restricting which related rows load.
+    pub filter: Option<Query>,
```
carllerche: Should this just be an `Expr`? Why is it a full query?

crates/toasty-macros/src/model/expand/fields.rs
```diff
@@ -163,9 +167,47 @@ impl Expand<'_> {
...
+        quote! {
+            /// Restrict which related rows preload when this path is passed
+            /// to `.include(...)`. The predicate is evaluated in the relation
+            /// target's scope; a filtered `HasOne`/`BelongsTo` loads `None`
+            /// when the related row does not match. Repeated includes of the
+            /// same relation AND their predicates.
+            #vis fn filter(self, predicate: #toasty::stmt::Expr<bool>) -> #toasty::stmt::Include<__Origin, #target_ty> {
+                #toasty::stmt::Include::from_path_and_filter(
+                    self.path,
+                    #toasty::stmt::Query::<#toasty::stmt::List<#model_ident>>::all().filter(predicate),
```
carllerche, line 201: Assuming you agree `stmt::Include` should take an expr and not a full Query, this gets simpler.

carllerche (dismissed):
Generally looks good. I left some feedback. The main point is I think stacking include should `OR` not `AND` given the terminology. I didn't do a deep review of the lowering code, I will do it after ^ is applied.

crates/toasty-core/src/stmt/include.rs
joeydewaal: I went with `Query` since I would like to add `order_by` and `limit` in the future. Also it preserves the target model's scope which we need when simplifying. (Similar to how `Path::any` is implemented.)

joeydewaal (commented):
Okay I've been thinking about this a bit.

I agree with switching to ` OR ` when joining multiple `include` calls. My reasoning for using `AND` was for supporting optional filters in for example webapps but we can support this in a different way:

```rust
// One include request
.include(user.posts().filter(a).filter(b)) // a AND b

// Two independent include requests; their requested rows are combined.
.include(user.posts().filter(a))
.include(user.posts().filter(b))           // a OR b
```

An unfiltered include requests all rows for that relation, so it overrides a filtered include for the same path:
```rust
.include(user.posts().filter(post_filter))
.include(user.posts())
```

This loads all posts because the two requests are combined as post_filter OR true.

This differs from a nested path:
```rust
.include(user.posts().filter(post_filter))
.include(
    user.posts()
        .comments()
        .filter(comment_filter),
)
```
The second include ends at comments, so it does not create an unfiltered include for posts. Only posts matching post_filter are loaded.

joeydewaal:
Okay I've applied your suggestions, cleaned up the pr a bit and fixed the tests. This is ready for another review.

carllerche:
@codex review this

crates/toasty-core/src/stmt/include.rs
```diff
@@ -0,0 +1,42 @@
+use super::{Node, Path, Query, Visit, VisitMut};
+
+/// An association preload entry on a `Returning::Model` clause.
+#[derive(Debug, Clone, PartialEq)]
+pub struct Include {
+    /// The relation path to preload.
+    pub path: Path,
+
+    /// Query applied to the related rows.
+    pub query: Query,
```
carllerche: I think `Option` would be better here. `Query::unit()` implies a query that returns no values (`unit()` might be a bad name for it, incidentally).

carllerche (commented):
One minor suggestion. Also, you can look at the automated codex review and see if anything useful is there.

carllerche:
Maybe think a bit about what stacking `.include(...)` on the same path means w/ query. E.g. what if you specify two different order by / limit? Those can't really stack.

crates/toasty/src/engine/lower/include.rs
```diff
@@ -320,7 +336,18 @@ impl LowerStatement<'_, '_> {
...
             }
-            return self.build_via_include_subquery(field_index, via, nested);
+            // `via` lowering does not thread per-relation filters through its
+            // JOIN chain yet; reject rather than silently drop them.
+            if top_filter.is_some()
+                || nested
+                    .iter()
+                    .any(|fi| query_filter_expr(&fi.query).is_some())
+            {
+                todo!("`.filter(...)` on a multi-step `via` relation include is not yet supported");
```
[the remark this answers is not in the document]
joeydewaal: Filter support on a multi-step via is for a follow up pr, so panicking for now is fine IMO.

crates/toasty/src/engine/verify.rs
```diff
@@ -308,6 +310,47 @@ impl Verify<'_, '_> {
...
+    /// loaded value has no way to represent a non-matching row, so the whole
+    /// query would fail with `RecordNotFound` depending on the data.
+    ///
+    /// Variant-rooted paths (relations inside embedded enums) are not
+    /// resolvable here and pass through unchecked.
+    fn verify_include_filters(&mut self, i: &stmt::Select) {
+        for include in i.returning.model_includes() {
+            let has_filter = match &include.query.body {
+                stmt::ExprSet::Select(select) => select.filter.expr.is_some(),
+                _ => false,
```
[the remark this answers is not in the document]
joeydewaal: Since filtering on a required 1-1 association is not allowed, I'd rather be a bit more conservative and also error when a bare include is called. User just shouldn't be calling the `filter` method here.

crates/toasty-core/src/stmt/include.rs
joeydewaal: I've made `Query` optional

PR #1092, opened 2026-07-15
fix(postgres): store Vec<native-enum> as a native enum array on Postgres

joeydewaal:
## Summary
Custom enum types in arrays in postgres used to be stored as `text[]` instead of `<native_enum>[]`.

## Type of change

- [ ] Other (please explain):

## Checklist

- [x] `cargo test` passes

## Notes for reviewers
N/A


crates/toasty/src/schema/embed.rs
```diff
@@ -38,4 +38,10 @@ pub trait Embed {
     {
         Path::field_at(Self::id(), index)
     }
+
+    /// The storage type this embed pins as a `Vec<T>` element: `Some` only for
+    /// a native unit enum, so the collection stores as `myenum[]`.
+    fn element_storage_ty() -> Option<toasty_core::schema::db::Type> {
```
carllerche: I don't think this new trait method is needed. I believe that the storage type, when left unspecified, can be figured out in `fn field_ty`

carllerche (changes requested):

crates/toasty/src/schema/embed.rs
joeydewaal: yes looks like `field_ty` is enough.

carllerche (approved):
ty.

PR #1093, opened 2026-07-15
feat(engine): bind multi-row INSERT as unnest arrays on PostgreSQL

### LINKED ISSUE #813: Use UNNEST/array binds for IN lists and multi-row inserts
### Problem

Statements with list-shaped operands currently emit one bind parameter per scalar value. Two common shapes hit this:

- `WHERE col IN (...)` and `NOT IN (...)`: one bind per list element.
- Multi-row `INSERT ... VALUES (...), (...), ...`: one bind per (row × column).

The cost shows up several ways:

- PostgreSQL caps a single statement at 65535 bind parameters. Bulk lookups and bulk inserts hit this limit on real workloads.
- The SQL string varies with input size, defeating prepared-statement reuse and bloating the planner cache. Association loads and batch inserts are the worst offenders, since Toasty issues both with input-dependent sizes.
- Query planning cost grows with the number of placeholders.

### Proposed solution

When the driver supports it, encode list-shaped binds as one array parameter per logical list and rewrite the SQL to consume it.

**IN-list and NOT IN.** PostgreSQL takes an array directly via `ANY` and `ALL`:

```sql
-- before
WHERE id IN ($1, $2, $3, ..., $N)

-- after
WHERE id = ANY($1)
```

```sql
-- before
WHERE id NOT IN ($1, $2, $3, ..., $N)

-- after
WHERE id <> ALL($1)
```

One bind regardless of list length.

**Multi-row INSERT.** PostgreSQL accepts parallel arrays through `unnest`:

```sql
-- before: N rows × C columns = N*C binds; query string varies with N
INSERT INTO users (id, name, email) VALUES
  ($1, $2, $3),
  ($4, $5, $6),
  ...;

-- after: C binds; query string fixed
INSERT INTO users (id, name, email)
SELECT * FROM unnest($1::bigint[], $2::text[], $3::text[]);
```

Bind count drops from `rows × columns` to `columns`. The query string no longer depends on row count, so the prepared statement caches across batch sizes.

The same shape extends to multi-row UPDATE via a join against `unnest`, but that can land later.

Equivalents on other databases:

- MySQL: no native `UNNEST`. Closest options are `JSON_TABLE($1, ...)` (8.0+) or falling back to per-element binds.
- SQLite: `json_each($1)` over a JSON array bind.
- DynamoDB: not applicable (no SQL).

The capability lives on the driver (`Capability`) and the SQL serializer consumes it. Drivers that don't advertise it keep the current per-element binding.

### Alternatives considered

- **Chunking**: split large `IN` lists or `INSERT` batches into multiple statements client-side. Avoids the bind-count limit but not planner-cache churn or round-trip count, and complicates result handling.
- **Temp tables**: `CREATE TEMP TABLE` + bulk load + join. Fastest for very large lists but adds round-trips and only pays off past a threshold.
- **Fixed-size prepared statements**: round list length up to a few canonical sizes (1, 10, 100, 1000) and pad. Reuses plans but wastes binds and still hits the cap.

The array and `unnest` approach is the cheapest and most general where supported, so it's the proposed default.

### Scope

Two stages, each landable on its own:

1. **IN and NOT IN** via `= ANY($1)` and `<> ALL($1)`. Touches the engine's bind extraction, `toasty-sql` serialization for `Expr::InList`, the driver `Capability` surface, and PostgreSQL `Value::List` binding.
2. **Multi-row INSERT** via `unnest`. Builds on the same `Type::List` and `Value::List` plumbing but adds a serializer path for `INSERT ... SELECT * FROM unnest(...)` and per-column array typing in the PostgreSQL driver.

SQLite (`json_each`) and MySQL (`JSON_TABLE`) can land separately as opt-in capabilities.

### Additional context

- PostgreSQL bind-count limit: https://www.postgresql.org/docs/current/protocol-message-formats.html (Bind message uses `int16` for parameter count).
- sqlx and Diesel both expose `= ANY($1)` patterns for the same reason.

joeydewaal:
## Summary
This pr rewrites inserts as unnests when supported. `INSERT ... VALUES (a, b), (c, d), ...` is rewritten to `INSERT ... SELECT * FROM unnest($1::t[], $2::t[])`.

Advantages:
* We no longer constrained by the max amount of bind parameters (65535).
* Bulk inserts with different amount of rows only emit one query plan.
* Better performance, the more values the longer it takes to build the query. Only having an bind parameter per column is more efficient.

## Type of change

- [ ] Other (please explain):

## Checklist

- [x] `cargo test` passes

## Notes for reviewers

Closes #813


carllerche (commented):
What do you think the pros/cons are of adding an `unnest` field to `Values` vs. explicitly adding an `unnest` function (`ExprFunc`). Long term, we want to add understanding for all these more edge-case SQL features so the engine can understand it, but that doesn't mean we. have to do it now.

carllerche:
> What do you think the pros/cons are of adding an `unnest` field to `Values` vs. explicitly adding an `unnest` function (`ExprFunc`). Long term, we want to add understanding for all these more edge-case SQL features so the engine can understand it, but that doesn't mean we. have to do it now.

@codex what do you think?

joeydewaal:
Because currently the engine doesn't really benefit from knowing about `unnest` and this is the simplest implementation. Though you bring up a good point. Also `unnest` can be called outside of this optimization in the future.

I'll push `unnest` into the engine.

crates/toasty-driver-integration-suite/src/tests/crud_batch_create.rs
```diff
@@ -68,6 +68,37 @@ pub async fn batch_create_many(test: &mut Test) -> Result<()> {
     Ok(())
 }
 
+/// On PostgreSQL this exercises the INSERT → `unnest` transpose with a NULL
+/// cell inside a column array bind.
+#[driver_test(id(ID))]
+pub async fn batch_create_with_null_field(test: &mut Test) -> Result<()> {
```
carllerche, line 74: If the specific SQL statement sent to the driver matters, that should be tested with the "exec log" tests (there are others that do so).

carllerche (changes requested):
I took a stab myself, building off of your PR here: https://github.com/tokio-rs/toasty/pull/1103/

The main thing is it gets rid of tracking types in the statement as that doe not reflect the AST of a SQL statement with `unnest`. You can tell me if you like it.

Regardless, since the PR's "feature" is SQL sent to the driver, we should test that it actually works using the `exec_log` style tests.

joeydewaal:
It makes sense to get rid of the types in `unnest` but I don't understand the advantage of using `FROM ROWS ...` instead of a plain `unnest` call. We could omit the types in the bind params like `FROM unnest($1, $2)` and just bind the types in the driver. Though I'm not sure whats better.

joeydewaal:
closing this in favor of #1103

Issue #422, opened 2026-03-09
Missing Use Case: Upsert


pikaju:
## Overview

Upsert — inserting a row or updating it if it already exists — is one of the most common write patterns in application development. PostgreSQL supports this via `INSERT ... ON CONFLICT DO UPDATE`, MySQL via `INSERT ... ON DUPLICATE KEY UPDATE`, and SQLite via `INSERT OR REPLACE` / `ON CONFLICT`. Toasty has no upsert support; inserts that hit a unique constraint simply fail.

### 1. Idempotent Event Ingestion

Processing webhooks, message queues, or API retries where the same event may arrive more than once.

```sql
INSERT INTO events (external_id, payload, received_at)
VALUES ('evt_abc123', '{"type":"payment"}', NOW())
ON CONFLICT (external_id) DO UPDATE
SET payload = EXCLUDED.payload, received_at = EXCLUDED.received_at;
```

Without upsert, the application must catch the unique-constraint error and issue a separate `UPDATE`, introducing a race condition window.

### 2. Counters and Accumulators

Incrementing a counter or appending to a value, creating the row on first encounter.

```sql
INSERT INTO page_views (url, view_count)
VALUES ('/home', 1)
ON CONFLICT (url) DO UPDATE
SET view_count = page_views.view_count + 1;
```

### 3. Last-Seen / Heartbeat Tracking

Updating a timestamp each time a device or user checks in, creating the record if it's the first time.

```sql
INSERT INTO device_heartbeats (device_id, last_seen_at, firmware_version)
VALUES ('d-42', NOW(), '3.1.0')
ON CONFLICT (device_id) DO UPDATE
SET last_seen_at = EXCLUDED.last_seen_at,
    firmware_version = EXCLUDED.firmware_version;
```

### 4. User Settings / Preferences

Saving a user preference that may or may not exist yet — a natural fit for `ON CONFLICT`.

```sql
INSERT INTO user_settings (user_id, key, value)
VALUES (7, 'theme', 'dark')
ON CONFLICT (user_id, key) DO UPDATE
SET value = EXCLUDED.value;
```

## Current Limitation in Toasty

- No `ON CONFLICT` or `DO UPDATE` representation in the statement AST.
- No `.upsert()` or `.on_conflict()` method on create builders.
- Inserts that violate a unique constraint return an error with no recovery path within the ORM.

## Possible Directions

[...]

PR #1094, opened 2026-07-16
docs: add update-returning design

joeydewaal:
## Summary
Adds a design doc for returning models after updating.

## Type of change

- [ ] Other (please explain):

## Checklist

- [x] `cargo test` passes

### DIFF
--- a/docs/dev/README.md
+++ b/docs/dev/README.md
@@ -33,6 +33,7 @@ Guide-level design documents for specific features. Use
 - [Optimistic Concurrency with `#[version]`](./design/field-version.md)
 - [`query!` Macro](./design/query-macro.md)
 - [Static Assertions for `create!` Required Fields](./design/static-assertions-create-macro.md)
+- [Returning Models from Updates](./design/update-returning.md)
 - [Upsert](./design/upsert.md)
 
 ## Roadmap
--- a/docs/dev/design/update-returning.md
+++ b/docs/dev/design/update-returning.md
@@ -2,172 +2,171 @@
 
 ## Summary
 
-Query-based updates gain `.all()`, `.first()`, and `.one()` modifiers that make
-`exec` return stored models. The builders compose with `toasty::batch()`.
-Backends without a native update-returning operation reject them before writing.
+Query updates keep returning `()`. `.affected_count()` returns the backend's
+affected count. `.all()`, `.first()`, and `.one()` update every matching row but
+control how many updated models are returned. They can select old or new values.
 
 ## Motivation
 
-Query-based updates currently discard database-generated values. Reading them
-requires a second query, which adds a round trip, can observe another writer's
-changes, and may fail when the update changes a filtered field.
+`()` is the only portable default. Cassandra, for example, does not report
+affected rows. Explicit counts avoid returning rows where supported.
 
-```rust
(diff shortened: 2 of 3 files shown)


carllerche (commented):
There are a few challenges to consider. 

- *what* does update return. It looks like you went with the post-update value, but another valid option is to return the *pre* update model. There are cases where you want to do one or the other. We should be able to have a way to configure *what* is returned. We can pick a default that is easier to write and have a way to configure it.

- Should the update by query *default* be to return something?

Here are my thoughts off the top of my head:

The *default* update should return the *new* values. This is a change from the current behavior.

Then, everything else is chained off of `Query`, so we have:

```rust
query.update(); // -> update builder configured to return new values
query.update_returning(New | Old | Count); // Count is the current behavior, kind of, just returns the count of impacted records.
```

I am not 100% sure what the full implications of this direction would be though, but ideally the update query has a typed return type so it can be combined with/ batch like you showed in this example.

joeydewaal:
So in the updated version of the doc I chose to keep `()` as the default return type.

With `returning_old()` and `returning_new()` it's possible to get the old/new values. These can only be used after selecting the cardinality with `first()`/`one()`/`all()`. `returning_new()` is the default since this is the most common use-case and most supported feature in the db's.

carllerche:
@joeydewaal why did you choose to keep `()` as the default?

carllerche:
The return configuration should also support returning the affected-row count. The conditional upsert use case in [#422](https://github.com/tokio-rs/toasty/issues/422#issuecomment-4624003708) needs this to distinguish acquiring a lease (`1`) from the `DO UPDATE ... WHERE` predicate rejecting the update (`0`). Since this design introduces configuration for mutation return values, it should account for affected-row counts as well.

joeydewaal:
I should've mentioned with the current design you can call `affected_count().exec(...)` to return the affected count.

`()` is the default because cassandra/scylladb (and redis) don't support returning the affected count by default. So if we/others want to support those db's we'd have to make a breaking change again. Making it opt in rather than the default is fine because I think that using the affected count is not that common.

carllerche:
> I should've mentioned with the current design you can call `affected_count().exec(...)` to return the affected count.
> 
> `()` is the default because cassandra/scylladb (and redis) don't support returning the affected count by default. So if we/others want to support those db's we'd have to make a breaking change again. Making it opt in rather than the default is fine because I think that using the affected count is not that common.

Do those support updating w/ anything except a key? What happens if you try to update a key that is missing?

joeydewaal:
Yeah its key only and an upsert for both so I guess the affected count is always 1.

carllerche (changes requested):
Let's rename the methods that control the return type to something like `returning_*`. Just having `affected_count` `all` and `one` in context of an update builder is confusing IMO as in it isn't clear it applies to the return type.

Also, maybe instead of `all` we can have `returning_new` and `returning_old` or something like that. I am not sure exactly what to do about `.one()` yet...

carllerche:
I guess we could do:

```rust
.returning_all() // -> Vec<Model> new values
.returning_one() // -> Model new value
.returning_all_old() // -> Vec<Model> old values
.returning_one_old() // -> Model new old value
```

PR #1095, opened 2026-07-17
feat: embed migrations in application binaries

Caisin:
## Summary

[Design](https://github.com/Caisin/toasty/blob/feat/embedded-migrations/docs/dev/design/embedded-migrations.md)

Add `embed_migrations!` to validate generated migration histories and embed their SQL files at compile time. Apply pending migrations through the target `Db` connection pool and report applied and skipped counts.

Document how applications bind separate migration sets to multiple data sources.

## Type of change

- [x] Other: adds a public API, its design document, implementation, and tests in one PR

## Checklist

- [x] `cargo test` passes
- [x] Touches a public API -> a design doc under `docs/dev/design/` describes the behavior

## Notes for reviewers

Focus on compile-time path validation in the proc macro and routing migration operations through the pooled connection. The change reuses existing driver migration methods and does not modify the `Driver` trait.


Caisin:
Hi @carllerche, all CI checks are passing. Could you please give it a quick review when you get a moment?

crates/toasty-macros/src/embed_migrations.rs
```diff
@@ -0,0 +1,134 @@
...
+        }
+        if !names.insert(entry.name.clone()) {
+            return Err(syn::Error::new(
+                path.span(),
+                format!("duplicate migration name {}", entry.name),
+            ));
+        }
+
+        let sql_path = migrations_dir.join(&entry.name);
+        read_utf8(&sql_path, &path, "migration SQL")?;
```
joeydewaal: This should probably use `std::fs::exists` instead of reading and discarding the data.

crates/toasty-macros/src/embed_migrations.rs
```diff
@@ -0,0 +1,134 @@
...
+        let id = entry.id;
+        migrations.push(quote! {
+            toasty::migration::MigrationFile::new(#id, #name, include_str!(#sql_path))
+        });
+    }
+
+    let history_path = path_literal(&history_path, path.span())?;
+    Ok(quote! {{
+        let _ = include_str!(#history_path);
+        toasty::migration::MigrationSet::new([#(#migrations),*])
```
joeydewaal: Given that `new` isn't const people can't store the embedded migrations in a static/const like:

```rust
static MIGRATIONS: MigrationSet = toasty::embed_migrations!();
```

crates/toasty-macros/src/embed_migrations.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+#[derive(Deserialize)]
+struct HistoryEntry {
+    id: u64,
+    name: String,
+}
+
+pub(crate) fn generate(input: TokenStream) -> syn::Result<TokenStream> {
+    let path = syn::parse2::<LitStr>(input)?;
```
joeydewaal: Since the cli generates migrations in the `toasty` directory by default we should probably have that as a default for when there's no path specified.

e.g.:
```rust
let migrations = embed_migrations!(); // -> read from "toasty/"
```

joeydewaal (commented):
Left some comments, let me know that you think. Generally I like the feature and think we should support something like this.

Caisin:
@joeydewaal  Thanks for the review. I pushed a follow-up commit addressing all three points:

  - `embed_migrations!()` now defaults to the `toasty/` directory.
  - The macro now uses `std::fs::exists` to validate referenced SQL files.
  - `MigrationFile::new` and `MigrationSet::new` are `const fn`, and the macro expansion now
  works in both `static` items and local bindings.

  I also added coverage for static initialization and the default directory behavior, and updated
  the docs.

crates/toasty-macros/src/embed_migrations.rs
```diff
@@ -0,0 +1,157 @@
...
+struct History {
+    version: u32,
+    migrations: Vec<HistoryEntry>,
+}
+
+#[derive(Deserialize)]
+struct HistoryEntry {
+    id: u64,
+    name: String,
+}
```
joeydewaal: I should've noticed this in the previous review but there are duplicated right? Can you figure out a way to deduplicate this. We should probably move these into `toasty-core`.

Caisin: @joeydewaal 
> Thanks, updated. I moved the migration history schema into toasty-core::migration and now
  > embed_migrations! reuses toasty_core::migration::History instead of maintaining a separate
  > parser/schema in toasty-macros. toasty::migration::{History, HistoryEntry} still re-exports
  > the same types, so the public API remains unchanged. I also added the missing snapshot_name
  > in the embedded test fixture so it uses the canonical history schema.

joeydewaal (approved):
Thanks!

carllerche (approved):

carllerche:
I'm not sure what is going on, but GitHub isn't letting me merge. Might be a Github service issue I guess. I will try again later.

Caisin:
> I'm not sure what is going on, but GitHub isn't letting me merge. Might be a Github service issue I guess. I will try again later.

Thanks for letting me know! No rush at all. Feel free to merge whenever GitHub is back to normal.

PR #1096, opened 2026-07-18
feat: support schema-less JSON payload storage

### LINKED ISSUE #672: feat: Postgresql JSONB type
## Problem Statement

Toasty does not currently support the JSONB Postgresql type. Many customers use JSONB for performance improvements - to the point where it's a blocker for some production use cases (anecdotally). Because of the downsides (slower writes), it's fine if we don't default to JSONB but we should at least give the user the option.

## Discussion

I can see the implementation going down 2 routes:

1. Extending `#[column]`:

```
struct User {
    #[key]
    #[auto]
    id: ID,

    #[serialize(json)]
    #[column(type = jsonb)]
    payload: Vec<String>,
}
```

2. Extending `#[serialize]`:

```
struct User {
    #[key]
    #[auto]
    id: ID,

    #[serialize(jsonb)]
    payload: Vec<String>,
}
```

I'm inclined towards the latter `#[serialize(jsonb)]`.

Caisin:
## Summary

- Add `toasty::JsonValue` for schema-less JSON payloads stored in native document columns.
- Preserve objects, arrays, scalars, explicit JSON null, and transparent serde conversion.
- Keep the existing string-backed `Json<T>` behavior unchanged and defer JSON path operations.

Closes #672

## Type of change

- [x] Other: adds the roadmap entry, design document, implementation, and tests in one PR.

## Checklist

- [x] `cargo test` passes
- [x] Touches a public API -> a design doc under `docs/dev/design/` describes the behavior
- [x] Touches the `Driver` trait or driver-observable behavior -> the design doc covers it

## Notes for reviewers

Please focus on the `Value::Json` marker that distinguishes database NULL from JSON null and on the cross-driver document encoding. PostgreSQL `jsonb` and MySQL `JSON` behavior were verified against local database instances. SQLite and Turso runtime integration tests pass, as does the full workspace suite. DynamoDB compiles and passes Clippy but was not run against a live service.

### DIFF
--- a/crates/toasty-core/src/schema/builder.rs
+++ b/crates/toasty-core/src/schema/builder.rs
@@ -291,13 +291,14 @@ impl BuildSchema<'_> {
                     })
                 };
 
-                let is_document = document_embed_id(&primitive.ty).is_some();
+                let is_document =
+                    document_embed_id(&primitive.ty).is_some() || primitive.ty.is_json();
 
                 if is_document {
                     if !self.db.document_collections {
                         return Err(crate::Error::unsupported_feature(format!(
-                            "model `{model_name}` field `{}` uses `#[document]` storage, \
-                             but this backend does not yet support `#[document]` fields.",
+                            "model `{model_name}` field `{}` uses document storage, \
+                             but this backend does not support document fields.",
                             field_name()
                         )));
                     }
--- a/crates/toasty-core/src/schema/db/ty.rs
+++ b/crates/toasty-core/src/schema/db/ty.rs
@@ -215,7 +215,7 @@ impl Type {
                 // `Type::list` collapses a list-of-documents back to one
                 // document, so a `#[document]` collection (`List(Model)`) lands
                 // here as a plain `Document`.
-                stmt::Type::Model(_) => Ok(Type::Document { binary: true }),
+                stmt::Type::Model(_) | stmt::Type::Json => Ok(Type::Document { binary: true }),
                 stmt::Type::List(elem) => Ok(Type::list(Self::from_app(elem, None, db)?)),
                 _ => Err(crate::Error::unsupported_feature(format!(
                     "type {:?} is not supported by this database",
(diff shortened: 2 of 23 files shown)


joeydewaal:
I think ideally we should support `#[column(type = json/jsonb)]` for `toasty::Json`. That way you can use `toasty::Json<serde_json::Value>` as struct field for schemaless JSON values and get to use the native json/jsonb storage type. Then move away from using text as the default storage type and favour jsonb when possible. What do you think?

Caisin:
> I think ideally we should support `#[column(type = json/jsonb)]` for `toasty::Json`. That way you can use `toasty::Json<serde_json::Value>` as struct field for schemaless JSON values and get to use the native json/jsonb storage type. Then move away from using text as the default storage type and favour jsonb when possible. What do you think?

this is better,when i need support multi db

Caisin:
Hi @carllerche, all CI checks are passing. Could you please give it a quick review when you get a moment?

carllerche:
Sorry for the delay. Getting to this now. I had a backlog of work to get through!

carllerche:
@codex review this

Caisin:
> Sorry for the delay. Getting to this now. I had a backlog of work to get through!

No worries at all! Thanks for taking the time, and good luck with the workload!

carllerche:
Right, this is interesting. I do agree with @joeydewaal that we *probably* don't want to add a new special toasty type and reusing `serde_json::Value` is probably best, but the question is *how*. We probably could just get it to work "out of the box" by implementing the right traits, but then we have a UX problem where we have `toasty::Json<T>` and `serde_json::Value` which do different things.

Another question I have is, should `toasty::Json<T>` default to using native json storage? If we do that it might make it more complicated to serialize to and from the serde value... I'm not 100% sure about that.

carllerche:
@Caisin please pause work on this PR. I 100% agree we need this. I think we need to think through some other changes to make this work better. I will be trying to push up more thoughts tomorrow.

Caisin:
Sounds great! Thanks for thinking through the design. I'll pause work on this for now and look forward to your update tomorrow!

carllerche:
I believe #1116 handles it. Thanks for the original PR and getting the ball rolling.

PR #1106, opened 2026-07-22
feat!: require explicit column types for JSON fields

carllerche:
## Summary

- Require every `Json<T>` field to declare `#[column(type = ...)]`.
- Propagate the requirement through field wrappers and support text-backed JSON compatibility.
- Update documentation, examples, integration coverage, and UI tests.

## Testing

- Not run (not requested)

### DIFF
--- a/crates/toasty-driver-integration-suite/src/scenarios/deferred_json_document.rs
+++ b/crates/toasty-driver-integration-suite/src/scenarios/deferred_json_document.rs
@@ -17,6 +17,7 @@ scenario! {
         id: ID,
 
         name: String,
+        #[column(type = text)]
         payload: toasty::Deferred<toasty::Json<Payload>>,
     }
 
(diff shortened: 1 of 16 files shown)


carllerche:
@joeydewaal @Caisin this is the first step towards supporting field support for `toasty::Json<serde_json::Value>`. We need to migrate `toasty::Json` to support db-native JSON columns. Doing this is a breaking change, so let's start by requiring an explicit column annotation. Follow up PR will add support for native json to `toasty::Json`.

joeydewaal:
Sounds like a good idea. I wonder if we could make a plain
 ```rust
#[column(type = json)]
field: serde_json::Value
```
  work, that would improve UX. Though I haven't looked into this.

carllerche:
@joeydewaal I think we could. I will definitely explore that option too.

PR #1110, opened 2026-07-22
fix(engine): skip transactions for read-only plans

joeydewaal:
## Summary
Currently when a plan is more than 1 db op the whole plan is wrapped in a transaction.
With this pr we only wrap plans with more than 1 mutating db op in a transaction. 

## Type of change

- [ ] Other (please explain):

## Checklist

- [x] `cargo test` passes

## Notes for reviewers
N/A


carllerche:
@codex review this.

carllerche:
@joeydewaal I don't think we want to land this as is. There are a few different dimensions. Mutation is one, but there is also the need to have consistent snapshots. We want consistent snapshots when loading includes. So, what we probably want is a transaction but maybe we want to set "repeatable read"?

joeydewaal:
So my biggest issue with this is that we have to do 2 extra round trips to the db (BEGIN, COMMIT) which costs on perf. A query with 1 relation included means 4 roundtrips, which is unfortunate.

A better solution for this is to optimize the generated plan to use joins when we can. This would cut down the amount of db ops to potentially 1 so we can skip the transaction wrapping.

I've kept the transaction wrapping for > 1 db ops and am using "repeatable read" as the default isolation level for read only plans.

carllerche:
With Postgres, we can leverage JSON apis to structure the response as a tree I think. Also, there is *probably* a way to reduce the round trips.

fredrikroos:
Postgres supports pipelining which makes BEGIN/COMMIT essentially free from a network POV. Also helps with the roundtrips for querying multiple relations. AFAIK the underlying pg client supports this, as long as you don't use the `Transaction` API, but instead just send BEGIN explicitly in SQL.

joeydewaal:
closing this as we should use other ways to reduce round trips like carl mentioned.

PR #1116, opened 2026-07-23
feat: support serde_json::Value fields

carllerche:
## Summary

Allow `serde_json::Value` to be used directly as a model field with an explicit `text`, `varchar`, `json`, or `jsonb` column type. The field reuses the existing `Json<T>` serialization path, so native JSON drivers continue to receive one serialized JSON value.

Add shared driver coverage for create, read, update, nested values, JSON null versus SQL `NULL`, PostgreSQL `json` and `jsonb`, and MySQL `json`. Document the direct field form in the derive documentation and user guide.

## Type of change

## Checklist

- [x] `cargo test` passes

## Notes for reviewers

No design doc is included because this maps `serde_json::Value` onto the existing `Json<T>` field behavior without adding a statement type or driver operation.

Validation includes the serde-enabled Toasty suite, macro doctests, shared-suite Clippy, PostgreSQL execution of the `json`, `jsonb`, and null cases, and MySQL execution of the `json` and null cases.

### DIFF
--- a/crates/toasty-driver-integration-suite/src/tests/type_serialize.rs
+++ b/crates/toasty-driver-integration-suite/src/tests/type_serialize.rs
@@ -348,6 +348,61 @@ pub async fn json_native_round_trip(t: &mut Test) -> Result<(), BoxError> {
 }
 
 #[driver_test(requires(native_json))]
+pub async fn json_value_native_round_trip(t: &mut Test) -> Result<(), BoxError> {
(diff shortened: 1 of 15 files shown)


carllerche:
@joeydewaal @Caisin this PR should support `serde_json::Value` as a field type while encoding it to the `jsonb` db column type.

joeydewaal (approved):

PR #1108, opened 2026-07-22
fix: serialize unconstrained numeric types in snapshots

joeydewaal:
## Summary
The column type repr of numeric (`Numeric(Option<(i32, i32)>)`) panics when trying to serialize `Numeric(None)` to toml. Toml can't represent this. Now we encode this as an array of either length 0 representing `Numeric(None)` or length 2 representing `Numeric(Some(..., ...))`.

## Type of change

- [x] Small / obvious change — bug fix, docs, internal cleanup, or test
- [ ] Implements a previously accepted design
      (link to the merged design doc under `docs/dev/design/`):
- [ ] Roadmap entry + design doc (no implementation in this PR)
- [ ] Other (please explain):

## Checklist

- [x] PR title uses [Conventional Commits](https://www.conventionalcommits.org/) format
- [x] `cargo fmt` and `cargo clippy` are clean
- [x] `cargo test` passes
- [x] Touches a public API → a design doc under `docs/dev/design/` describes the behavior
- [x] Touches the `Driver` trait or driver-observable behavior → the design doc covers it

## Notes for reviewers
N/A

Closes #1107

PR #1119, opened 2026-07-23
refactor: simplify numeric snapshot serialization

carllerche:
## Summary

- follow the narrower numeric snapshot serialization approach from #1108
- encode numeric parameters only as zero- or two-element arrays
- keep serialization fallible inside `Snapshot::save` without a separate public rendering API
- retain the focused snapshot tests and reporter-shaped CLI regression coverage from #1115

## Type of change

## Checklist

- [x] `cargo test` passes
- [ ] Touches a public API → no design doc; this removes the newly added `Snapshot::to_toml_string` API before release

## Notes for reviewers

This deliberately removes the legacy nullable decoder and restores the previous CLI write ordering. The regression tests added in #1115 remain unchanged and pass with the smaller implementation.

### DIFF
--- a/crates/toasty-cli/src/migration/generate.rs
+++ b/crates/toasty-cli/src/migration/generate.rs
@@ -295,7 +295,6 @@ impl GenerateCommand {
             checksum: None,
         });
 
-        let snapshot = generated.snapshot.to_toml_string()?;
         let migration = generated.migration;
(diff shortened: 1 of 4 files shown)


carllerche:
Eh, given the empty array issue which isn't true on `main` (I think), I'm just going to close this.

PR #1124, opened 2026-07-24
fix: paginate with multiple order by keys

### LINKED ISSUE #1121: Multi-column keyset pagination returns an empty next page when leading sort values are equal
### What you did

I paginated a SQLite-backed model using two ascending sort columns:

```rust
.order_by((Item::fields().group().asc(), Item::fields().id().asc()))
```

All three rows had the same value in the leading `group` column. I requested two rows per page, then passed the first page's `next_cursor` to the same query using `.after(...)`.

### What you expected

The first page should contain IDs `1` and `2`, and the second page should contain ID `3`.

### What actually happened

The first page contained IDs `1` and `2` and provided a `next_cursor`, but the second page was empty:

```text
assertion `left == right` failed:
multi-column pagination must compare cursor keys lexicographically

left:  []
right: [3]
```

### Minimal reproducer

```rust
use toasty::stmt::Page;

struct Item {
    #[key]
    id: i64,
    group: i64,
}

#[tokio::test]
async fn multi_column_keyset_pagination_advances_with_an_equal_leading_value() {
    let mut db = toasty::Db::builder()
        .models(toasty::models!(Item))
        .connect("sqlite::memory:")
        .await
        .unwrap();
    db.push_schema().await.unwrap();

    for id in 1..=3 {
        Item::create()
            .id(id)
            .group(1)
            .exec(&mut db)
            .await
            .unwrap();
    }

    let first_page: Page<Item> = Item::all()
        .order_by((
            Item::fields().group().asc(),
            Item::fields().id().asc(),
        ))
        .paginate(2)
        .exec(&mut db)
        .await
        .unwrap();

    assert_eq!(
        first_page
            .items
            .iter()
            .map(|item| item.id)
            .collect::<Vec<_>>(),
        [1, 2]
    );

    let next_cursor = first_page
        .next_cursor
        .expect("a full first page should provide a next cursor");

    let second_page: Page<Item> = Item::all()
        .order_by((
            Item::fields().group().asc(),
            Item::fields().id().asc(),
        ))
        .paginate(2)
        .after(next_cursor)
        .exec(&mut db)
        .await
        .unwrap();

    assert_eq!(
        second_page
            .items
            .iter()
            .map(|item| item.id)
            .collect::<Vec<_>>(),
        [3],
        "multi-column pagination must compare cursor keys lexicographically"
    );
}
```

### Database driver

SQLite

### Toasty version

`0.9.0`

### Additional context

The same behavior occurs when the two `order_by` calls are chained instead of passed as a tuple.

joeydewaal:
## Summary
Closes #1121

Keyset pagination over a multi-column `order_by` returned an empty next page when rows tied the cursor on a leading sort column. The SQL cursor rewrite (`rewrite_offset_after_as_filter`) translated a cursor `(group: 1, id: 2)` into a conjunction of strict per-column comparisons — `group > 1 AND id > 2` — which excludes rows such as `(group: 1, id: 3)` that tie on `group`.

The rewrite now emits the lexicographic expansion:
```sql
group > 1 OR (group = 1 AND id > 2)
```

## Type of change

- [ ] Other (please explain):

## Checklist

- [x] `cargo test` passes

## Notes for reviewers
N/A

### DIFF
--- a/crates/toasty-driver-integration-suite/src/tests.rs
+++ b/crates/toasty-driver-integration-suite/src/tests.rs
@@ -64,6 +64,7 @@ pub mod infra_reset_db;
 pub mod infra_sync_send;
 pub mod key_unsigned;
 pub mod lift_belongs_to_complex_filter;
+pub mod paginate_multi_column;
 pub mod query_count;
 pub mod query_in_list;
 pub mod raw_identifier_fields;
--- a/crates/toasty-driver-integration-suite/src/tests/paginate_multi_column.rs
+++ b/crates/toasty-driver-integration-suite/src/tests/paginate_multi_column.rs
@@ -1,0 +1,76 @@
+//! Test keyset pagination over a multi-column `order_by`.
+//!
+//! The cursor must be compared lexicographically: a row that ties the cursor
+//! on a leading sort column but is beyond it on a later column belongs on the
+//! next page. These are gated on `sql` because SQL drivers rewrite the cursor
+//! into a WHERE filter; NoSQL drivers use a driver-level cursor instead.
+
+use crate::prelude::*;
+use toasty::stmt::Page;
+
+#[driver_test(requires(sql))]
+pub async fn paginate_multi_column_equal_leading_values(test: &mut Test) -> Result<()> {
+    #[derive(Debug, toasty::Model)]
+    struct Item {
+        #[key]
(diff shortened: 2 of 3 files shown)


joeydewaal:
@codex review

carllerche (approved):

PR #1041, opened 2026-06-13
feat: add model comments

Caisin:
## Summary

Adds `#[comment("...")]` / `#[comment = "..."]` for root models and fields so schema comments can live with the model definition.

The metadata is stored in app and database schemas, included in schema diffs, and serialized during SQL migrations:

- PostgreSQL emits `COMMENT ON TABLE` / `COMMENT ON COLUMN` statements.
- MySQL emits inline `COMMENT` clauses on create and `ALTER TABLE ... COMMENT` / `CHANGE COLUMN ... COMMENT` on change.
- SQLite and Turso retain comments in Toasty's schema metadata but emit no comment DDL.

Includes a design doc, guide updates, SQL migration tests, and a user-level SQLite schema test for macro propagation.

## Type of change

- [x] Other (please explain): MVP user-facing schema metadata feature requested for table and column comments.

## Checklist

- [x] PR title uses [Conventional Commits](https://www.conventionalcommits.org/) format
- [x] `cargo fmt` and `cargo clippy` are clean
- [x] `cargo test` passes
- [x] Touches a public API → a design doc under `docs/dev/design/` describes the behavior
- [x] Touches the `Driver` trait or driver-observable behavior → the design doc covers it

## Notes for reviewers

Validation run locally:

- `cargo fmt`
- `cargo clippy -- -D warnings`
- `cargo clippy -p toasty-sql --all-targets -- -D warnings`
- `cargo test -p toasty-sql`
- `cargo test -p tests --test sqlite`
- `cargo check -p toasty-macros && cargo check -p toasty && cargo check -p toasty-driver-integration-suite`

I did not run live MySQL or PostgreSQL driver suites.

PR #1132, opened 2026-07-26
feat!: add database schema comments

Caisin:
## Summary

[Design document](https://github.com/Caisin/toasty/blob/feat/schema-comments/docs/dev/design/schema-comments.md)

Add `#[comment = "..."]` attributes for models and fields and carry those comments into database schemas.

- PostgreSQL and MySQL emit native table and column comments.
- `MigrationConfig::schema_comments(true)` opts generated migrations into managing comments; the default is `false`.
- Schema creation, alteration, rename, table reconstruction, `push_schema()`, and reset/replay preserve comments.
- SQLite, Turso, and DynamoDB ignore comments without emitting unsupported operations.

This supersedes #1041 with a current implementation and a complete design document. #1087 adds Rustdoc to generated model methods; it does not implement database schema comments.

## Type of change

- [x] Other: implementation and design document submitted together for review

## Checklist

- [x] `cargo test` passes

## Notes for reviewers

This is a breaking API change because public `Capability` and schema struct literals must initialize the new comment fields.

Verification:

- `cargo fmt`
- `cargo clippy --workspace --all-targets`
- `cargo test --workspace`
- Live PostgreSQL verification for `push_schema()` and comment-update migrations through the system catalogs
- Live MySQL verification for `push_schema()`, table-comment updates, and column alteration while preserving comments

The full PostgreSQL and MySQL driver integration suites were not run.


carllerche:
Thanks. I would rather not add a new top-level attribute. Since this applies to the schema, we could probably include it with existing `#[column]` and `#[table]`. e.g. `#[column(comment = "foo")]`.

Another option would be to add it in doc comments?

```rust
/// @db this line goes to the database schema
field: String,
```

Since doc comments are available in the macro, this should work. Only question is if it is a good idea.

carllerche:
@Caisin btw, are you on Discord? It might be easier to iterate on PR details there

Caisin:
Thanks for the feedback, @carllerche!
​Attribute Design: That makes total sense. I agree that attaching comment directly to the existing #[table(comment = "...")] and #[column(comment = "...")] is much cleaner than introducing a new top-level attribute. I'll refactor the PR to follow this approach.
​Discord: Yes, I am! Here is my Discord invite link: https://discord.gg/aaq8gr46 . Looking forward to catching up and iterating on the details there!

crates/toasty-cli/src/migration/generate.rs
```diff
@@ -256,9 +256,13 @@ impl GenerateCommand {
...
         let rename_hints = collect_rename_hints(&previous_schema, &schema)?;
-        let Some(generated) =
-            migration::generate(db.driver(), &previous_schema, &schema, &rename_hints)
-        else {
+        let Some(generated) = migration::generate_with_options(
+            db.driver(),
+            &previous_schema,
+            &schema,
+            &rename_hints,
+            migration::GenerateOptions::new().schema_comments(config.migration.schema_comments),
```
joeydewaal, line 264: Why are we passing in `GenerateOptions` here, this feels out of place.

crates/toasty-core/src/schema/builder/table.rs
```diff
@@ -157,7 +157,9 @@ impl BuildSchema<'_> {
 
         if !self.table_lookup.contains_key(&table_name) {
             let id = self.register_table(&table_name);
-            self.tables.push(Table::new(id, table_name.clone()));
+            let mut table = Table::new(id, table_name.clone());
+            table.comment = model.comment.clone();
+            self.tables.push(table);
```
joeydewaal, line 162: Setting the comment here feels a bit wrong, shouldn't we set this in `BuildTableFromModels::build`, wdyt?

joeydewaal:
@codex review

crates/toasty-sql/src/stmt/alter_column.rs
```diff
@@ -101,18 +101,30 @@ impl AlterColumnChanges {
...
     pub fn alter_column(
-        column: &Column,
+        previous: &Column,
+        next: &Column,
         changes: AlterColumnChanges,
         capability: &Capability,
     ) -> Self {
+        let mut column_def =
+            ColumnDef::from_schema(previous, &capability.storage_types, capability);
+        column_def.comment = next.comment.clone();
```
joeydewaal, line 124: Can't you put the comment in `AlterColumnChanges`?

joeydewaal (commented):
Left some comments, let me know what you think

joeydewaal:
@Caisin did you get a chance to look at my review?

PR #1154, opened 2026-08-06
feat(cli): ship a standalone toasty binary

carllerche:
## Summary

Implements [`docs/dev/design/standalone-cli.md`](https://github.com/tokio-rs/toasty/blob/worktree-bright-moseying-penguin/docs/dev/design/standalone-cli.md). The design doc is not merged yet — it lives on `worktree-bright-moseying-penguin`.

`cargo install toasty-cli` now provides a `toasty` command that runs from any Cargo package, replacing the per-project migration binary users had to write. No `Cargo.toml` changes, no call site.

```
toasty migrate generate --flavor postgresql --name init
toasty migrate apply --url postgres://localhost/app
toasty -p api migrate generate --flavor postgresql
```

Three pieces:

- **Schema extraction** (`toasty::schema::dump`). A `linktime` constructor, gated on `cfg(debug_assertions)` and a default-on `schema-dump` feature, walks the `inventory` registrations and writes the resolved `app::Schema` as JSON when `TOASTY_DUMP_SCHEMA` is set. The CLI builds the package's bin — or its lib as a `cdylib` via `cargo rustc --crate-type cdylib` — and runs it. The user's `main` never executes. Serializing the payload required serde on the `app::Schema` graph in `toasty-core`.
- **Flavor-keyed DDL.** `SqlFlavor` moves into `toasty-core` and is reported through `Capability::sql_flavor`, so `--flavor postgresql` renders PostgreSQL DDL with no PostgreSQL to connect to.
- **The binary** (`toasty-cli/src/standalone/`). `cargo metadata` for package selection, artifact discovery from `--message-format=json-render-diagnostics`, and the hidden `__load-cdylib` re-exec. The existing commands were refactored to take a driver or a schema instead of a `Db`, so this and `ToastyCli` share one implementation.

## Type of change

      (link to the merged design doc under `docs/dev/design/`): [standalone-cli.md](https://github.com/tokio-rs/toasty/blob/worktree-bright-moseying-penguin/docs/dev/design/standalone-cli.md) — **not yet merged**

## Checklist

- [x] `cargo test` passes (3335 tests)

## Notes for reviewers

**Breaking:** `Driver::generate_migration` is removed — drivers no longer render DDL. Callers use `toasty_sql::generate_migration(flavor, diff)` or `toasty::migration::generate`, which now takes a `Flavor`. `Capability` gains a `sql_flavor` field, validated against `sql`.

**Two deviations from the design, both worth a look:**

1. *The env var is set inside the `__load-cdylib` child, not inherited.* The design has the parent set `TOASTY_DUMP_SCHEMA` on the child's environment. That doesn't work: the CLI links `toasty` too, so its own constructor fired at startup and dumped an empty schema before ever reaching the `dlopen`. Setting it immediately before `Library::new` means the only constructor that sees it belongs to the library being loaded. This is the one real bug the design as written would have shipped.

2. *The payload has a marker line and a version envelope.* Other constructors run before ours and can print, so the CLI locates the payload by marker rather than assuming it owns stdout. The envelope carries the producing `toasty` version, so a CLI installed at a different version reports a mismatch instead of a serde error.

Smaller notes: `Flavor::generate_migration` is a free function, since an inherent impl isn't allowed on a type from another crate and the enum has to live in `toasty-core` for `Capability` to report it. The `toasty`-dependency precheck walks the resolve graph rather than direct dependencies — a binary whose models live in a sibling library never names `toasty` in its own manifest.

Both design open questions landed where the design leaned: the `cfg(debug_assertions)` gate stayed, and `--bin <name>` is a target-selection flag alongside `-p`.

**Coverage:** `crates/toasty-cli/tests/standalone_cli.rs` builds throwaway packages and runs the real binary over both extraction paths on every CI run (~29s), which is what keeps the constructor-survives-the-cdylib-link property honest across platforms.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -108,6 +108,12 @@ heck = "0.5.0"
 indexmap = "2.13.0"
 index_vec = "0.1.4"
 inventory = "0.3.24"
+libloading = "0.8.9"
+linktime = { version = "0.18.0", default-features = false, features = [
+	"std",
+	"ctor",
+	"proc_macro",
+] }
 # Must match the version used by `postgres-protocol` (the postgres driver
 # iterates over its `FallibleIterator` types); do not bump independently.
 fallible-iterator = "0.2"
--- a/crates/toasty-cli/Cargo.toml
+++ b/crates/toasty-cli/Cargo.toml
@@ -13,6 +13,27 @@ categories.workspace = true
 readme.workspace = true
 documentation = "https://docs.rs/toasty-cli"
 
+[features]
+default = ["cli"]
+
+# The standalone `toasty` binary. Pulls in the SQL drivers it connects with and
+# the machinery for building a user's package and reading the schema back out.
+# Depend on this crate with `default-features = false` to use it purely as a
+# library for a CLI of your own.
+cli = [
+	"dep:libloading",
+	"dep:serde_json",
+	"dep:tokio",
+	"toasty/mysql",
+	"toasty/postgresql",
+	"toasty/sqlite",
+]
+
+[[bin]]
+name = "toasty"
+path = "src/bin/toasty.rs"
+required-features = ["cli"]
+
 [dependencies]
 toasty = { workspace = true, features = ["migration"] }
 toasty-core.workspace = true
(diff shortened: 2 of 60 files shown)


carllerche:
#1157 takes precedence

PR #1170, opened 2026-08-08
feat: accept #[belongs_to] fields in embedded types

carllerche:
## Summary

Implements step 1 of the polymorphic-relations design: `#[belongs_to]` fields are accepted inside `#[derive(Embed)]` structs and enum variants.

The relation field registers its schema entry (target + foreign key resolved against sibling fields of the same struct/variant) but maps to **no columns** — the key fields own the storage and are indexed with the existing field-level / enum-level forms. The field must be `toasty::Deferred<..>` (enforced by a const assertion): its record slot encodes as `Null` on write and decodes to the unloaded state on read. Creating and updating supply the embed value with explicit keys, `match` reads the stored keys back, the owner loads with an ordinary `get_by_*`, and key fields stay queryable through the existing variant filter paths — discriminant-gated, so a `#[shared]` key column never cross-matches variants.

Not in this step (per the design's implementation order): `.include()`, setting the relation from a model value (`Owner::Human { human: &alice }`), and inverse `has_many` pairing. Field-level `#[shared]`/`#[index]` on a relation field and non-`Deferred` relation types are compile errors.

Also fixes a pre-existing engine bug this surfaced: filtering a variant field whose model type differs from its stored column type (e.g. `uuid::Uuid` stored as text) panicked in the SQL serializer, because the decode cast inside the enum's `Match` arm only unfolds during post-lower simplify, after the lowering-time cast rule has run. Simplify now strips the cast and converts the constant side (separate commit, with a no-relations regression test).

The branch also carries a small doc correction: the design doc listed enum-level non-unique `#[index(...)]` as unshipped step 1, but it shipped in #1078; the step list is renumbered.

## Type of change

      (link to the merged design doc under `docs/dev/design/`): [polymorphic-relations.md](https://github.com/tokio-rs/toasty/blob/main/docs/dev/design/polymorphic-relations.md)

## Checklist

- [x] `cargo test` passes (SQLite full suite; DynamoDB full suite against local DynamoDB)

## Notes for reviewers

- `mapping::Field::Relation` was already the "maps to no columns" mapping for root models; the builder change is only lifting the root-only `assert!(!force_nullable)` so it applies inside enum variants and nullable struct embeds.
- Foreign-key `source` indices for enum variants are remapped from variant-local to the embed's global field indices at parse time (`from_enum_ast`), since `BelongsTo::from_ast` resolves `key` against the variant's own field list.
- Relation fields keep their positional record slot everywhere (schema `FieldId`, encode, decode, reload) rather than being skipped — positional indexing is load-bearing in ~6 places, and root models already use the slot-holds-`Null` convention.
- Generated API surface deliberately omits what step 1 cannot support: no fields-struct accessors, no update-builder setters, and no include processing for embedded relation fields. `codegen_support::embedded_relation_expr` rejects a loaded `Deferred` at runtime with a message pointing at explicit keys.
- New suite `embed_relation.rs` (5 tests: enum + struct embeds, shared and per-variant keys, optional embed, variant-switch update, gated key filter); three trybuild UI tests cover the error paths.

https://claude.ai/code/session_01SmiM72YpE84pdWmp6Qxd2R

### DIFF
--- a/crates/toasty-core/src/schema/builder/table.rs
+++ b/crates/toasty-core/src/schema/builder/table.rs
@@ -951,7 +951,9 @@ impl<'a, 'b> MapField<'a, 'b> {
                 }
             }
             app::FieldTy::BelongsTo(_) | app::FieldTy::Has(_) | app::FieldTy::Via(_) => {
-                assert!(!self.force_nullable);
+                // A relation field maps to no column, so the nullability
+                // context (`force_nullable`, set inside enum variants and
+                // nullable struct embeds) has nothing to apply to.
                 let bit = self.build.next_bit();
                 Ok(mapping::Field::Relation(mapping::FieldRelation {
                     field_mask: stmt::PathFieldSet::from_iter([bit]),
--- a/crates/toasty-driver-integration-suite/src/tests.rs
+++ b/crates/toasty-driver-integration-suite/src/tests.rs
@@ -40,6 +40,7 @@ pub mod embed_enum_shared_type;
 pub mod embed_enum_string_discriminant;
 pub mod embed_enum_unit;
 pub mod embed_newtype;
+pub mod embed_relation;
 pub mod embed_struct;
 pub mod embed_struct_index;
 pub mod embed_struct_optional;
--- a/crates/toasty-driver-integration-suite/src/tests/embed_enum_filter_variant_field.rs
+++ b/crates/toasty-driver-integration-suite/src/tests/embed_enum_filter_variant_field.rs
@@ -172,3 +172,59 @@ pub async fn filter_variant_field_with_cast_storage(t: &mut Test) -> Result<()> 
 
     Ok(())
 }
+
+/// Same as above, but the variant field overrides its storage type away from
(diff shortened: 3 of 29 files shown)


carllerche:
Re the Codex comment on `expr_binary_op.rs`: confirmed and fixed in 5c9f8ac53 — the rule now resolves the referenced column and converts the constant to its actual stored type, bailing on anything that is not a physical column. Regression test `filter_variant_field_with_storage_override` covers the text-stored-UUID-on-SQLite case (it reproduces the silent zero-row match against the previous code).

Two related notes from investigating: the lowering-time analog of this rule (`lower_expr_binary_op`) has the same `native_type_for` shape for root-level fields — pre-existing and currently unreachable by tests, left untouched here. And filtering a variant field of a single-data-variant enum trips an unrelated pre-existing assert in `engine/verify.rs` (variant paths encode record positions, the verify pass resolves them as `[variant, global field]`); the regression test uses two variants to sidestep it.

carllerche:
@codex review this again

carllerche:
@codex review this again

PR #1172, opened 2026-08-09
feat: reference embedded relations with model values

carllerche:
Implement step 2 of the polymorphic-relations design: a relation stored in an embedded type can now be compared to and set from a model value.

Filters resolve a projection into an embed to the belongs_to it lands on or passes through, using the is_variant gates the typed layer fuses next to variant-rooted comparisons (enum record slots are variant-local and ambiguous without them). A comparison at the relation substitutes the sibling key slot(s); traversal into the target model and in_query lift to a foreign-key IN subquery. An enum decode Match on the LHS of a membership test is eliminated eagerly at lower time into variant-gated terms, since on non-SQL backends the subquery detaches into its own statement before post-lower simplification could reach it. The verify pass gains a record-slot-space resolver for expression projections; Schema::resolve is path-space and only accepted step-1 shapes by accident.

On the write side, generated IntoExpr bodies fill key slots from a loaded relation value by reading the referenced field off the parent model, and create!/update! rewrite variant literals into derive- generated expression builders, so the parent can be passed by reference with the key omitted: Owner::Human { human: &alice }. Non-relation builder setters are monomorphic in the declared field type so values keep type-checking exactly as in a plain struct literal.


crates/toasty-macros/src/model/expand/embedded_enum.rs
```diff
@@ -323,6 +334,185 @@ impl Expand<'_> {
...
+                        /// merged over `slots` (winning) at `into_expr`.
+                        rel_keys: Vec<#toasty::Option<#toasty::core::stmt::Expr>>,
+                    }
+
+                    impl #model_ident {
+                        #[doc(hidden)]
+                        #vis fn #entry_ident() -> #builder_ident {
+                            let mut slots = Vec::with_capacity(#slot_count);
+                            slots.push(#discriminant_expr);
+                            slots.resize_with(#slot_count, #toasty::core::stmt::Expr::null);
```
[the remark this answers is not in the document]
carllerche: Valid — the rewrite dropped Rust's exhaustiveness check and nothing downstream validates variant slots (root insert validation only checks root fields, and the enum record itself is non-null). Fixed in 5555b9904: the generated builder panics at expression-build time when a required slot is neither set explicitly nor filled from a loaded relation value, with distinct messages for plain fields and relation-fillable key fields. Covered by `tests/tests/embed_variant_literal_required.rs`.

crates/toasty/src/engine/lower/lift_in_subquery.rs
```diff
@@ -574,6 +622,150 @@ fn lift_relation_path_predicate(
...
+    if !e.op.is_eq() && !e.op.is_ne() {
+        return false;
+    }
+
+    for side in [&mut e.lhs, &mut e.rhs] {
+        if let Some(resolved) = resolve_embedded_relation(cx, side, gates)
+            && resolved.tail.is_empty()
+        {
+            **side = resolved.key_expr;
+            return true;
```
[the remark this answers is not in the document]
carllerche: Valid — the early return left the second relation operand lowering to its storage-less `Null` slot. Fixed in 63ceaa4bc: the rewrite now continues through both operands, and post-lower simplify gains a cast-vs-cast rule (mirroring the existing cast-vs-value rule) for the column-versus-column comparison this produces. Covered by `compare_embedded_relation_to_embedded_relation` in the integration suite.

crates/toasty-macros/src/model/expand/model.rs
```diff
@@ -465,44 +465,54 @@ impl Expand<'_> {
...
+            };
+
+            // A key field backing a sibling relation takes its value from
+            // the loaded parent model when one is present; the explicit key
+            // stands otherwise.
+            match key_fill.get(&field.id) {
+                Some((rel_ident, target_ident)) => quote!(
+                    match #toasty::embedded_relation_target(&self.#rel_ident) {
+                        #toasty::Option::Some(__rel) =>
+                            #toasty::into_untyped_expr::<#target_ty, _>(&__rel.#target_ident),
```
carllerche: [P2] Preserve relations to models with private key fields

Reading `__rel.#target_ident` directly makes the embed derive depend on the Rust visibility of the target field. A public `models::Human` with a private `id`, referenced by an embedded struct outside `models`, compiles on the merge base but fails here with E0616—even when every write supplies the key explicitly and leaves the relation unloaded. The enum key-fill paths introduce the same access. Extract the referenced value through code generated in the target model’s module or a model-level accessor so existing cross-module relations do not require exposing their fields.

crates/toasty/src/engine/lower/lift_in_subquery.rs
```diff
@@ -574,6 +622,155 @@ fn lift_relation_path_predicate(
...
+        return false;
+    }
+
+    let mut rewrote = false;
+
+    for side in [&mut e.lhs, &mut e.rhs] {
+        if let Some(resolved) = resolve_embedded_relation(cx, side, gates)
+            && resolved.tail.is_empty()
+        {
+            **side = resolved.key_expr;
```
carllerche: [P2] Resolve model comparisons through the configured referenced key

Replacing the relation with its source key is only sufficient when that key references the target’s primary key: `IntoExpr` for a model value emits its primary key. Using this PR’s `Owner::Bot` fixture (`references = serial`), `owner().bot().matches(|v| v.bot().eq(&bot))` compares the stored string serial with `bot.id`, a UUID. I reproduced a panic in SQLite’s value encoder; equal-typed but different keys can instead silently miss the row. For non-primary references, resolve the model’s primary-key predicate through a target subquery that returns the configured referenced field(s), rather than comparing the foreign key directly to the model expression.

crates/toasty/src/engine/lower/lift_in_subquery.rs
```diff
@@ -68,6 +72,29 @@ impl<'a> LiftInSubquery<'a> {
...
+        let lifted = match expr {
+            stmt::Expr::BinaryOp(e) => {
+                if rewrite_embedded_relation_operand(&self.cx, e, gates) {
+                    return;
+                }
+                lift_embedded_relation_comparison(&self.cx, e, gates)
+            }
+            stmt::Expr::Like(e) => lift_embedded_relation_like(&self.cx, e, gates),
+            stmt::Expr::InSubquery(e) => lift_embedded_relation_in_subquery(&self.cx, e, gates),
+            _ => None,
```
carllerche: [P2] Lower explicit list membership on embedded relation paths

This dispatcher leaves `Expr::InList` untouched, so the newly exposed relation paths still refer to their storage-less relation slot when used with the public list-membership API. In the existing Human/Animal fixture, `owner().human().matches(|v| toasty::stmt::Expr::in_list(v.human(), [&alice, &bea]))` compiles but panics in the SQLite serializer on an unlowered `ExprMatch`, instead of returning the two Human-owned rows. Rewrite the relation operand to its key for list membership too, and eliminate the enum decode around that membership before it reaches the driver.

carllerche (commented):
Found three issues to fix before merging: cross-module embedded relations can stop compiling, comparisons against a model value use the wrong key for non-primary references, and explicit relation list membership panics. Each finding was reproduced against this head.

Validation: all 801 existing SQLite tests, the three required-variant-slot tests, and the two cast-preservation tests pass. Two additional SQLite reproductions fail as described inline. The private-key derive example compiles against the PR merge base (7af9ecc50203ac8cb2abc7833dfa336670c78a4d) and fails with E0616 against this head. External database drivers were not run.

crates/toasty/src/engine/lower/lift_in_subquery.rs
carllerche: Valid, but not introduced here. The root-level `belongs_to` eq rewrite on `main` has the same behavior: it substitutes the foreign-key column while the model value lowers to its primary key, so a plain (non-embedded) relation with `references = serial` panics the same way. The embedded path in this PR inherits that. Tracked in #1219; out of scope for this PR.

crates/toasty/src/engine/lower.rs
```diff
@@ -2070,6 +2117,62 @@ impl LoweringContext<'_> {
...
+                _ => stmt::Expr::record(Vec::<stmt::Expr>::new()),
+            }
+        }
+        stmt::Expr::Value(stmt::Value::Record(record))
+            if record.fields.first() == Some(disc_value) =>
+        {
+            stmt::Expr::record(record.fields.into_iter().skip(1).map(stmt::Expr::Value))
+        }
+        stmt::Expr::Value(_) => stmt::Expr::error("value is not the selected variant"),
+        lowered => todo!("variant selection on lowered expression {lowered:#?}"),
```
carllerche, line 2204: [P2] Handle variant selection through an optional enclosing embed

For `Item { metadata: Option<Metadata> }` with `Metadata { contact: Contact }`, filtering on the nested enum payload now reaches this `todo!`. A typed `Contact` path built with `Field::new_path(Item::fields().metadata().chain(Metadata::fields().contact()))`, followed by `contact.email().address().eq("alice")`, passes on the base commit and panics here. The optional struct lowers to a presence-check `Match`, so the enum arrives as a projection through that match rather than as the direct enum decode handled above. Resolve the enclosing optional projection while preserving its null behavior before extracting the selected variant payload.

crates/toasty/src/engine/lower/lift_in_subquery.rs
```diff
@@ -557,21 +618,193 @@ fn lift_relation_path_predicate(
...
+
+/// `<relation-path> IN (list)`: substitute the projection of the relation's
+/// key field(s) for the relation reference, in place — the
+/// [`rewrite_embedded_relation_operand`] case for list membership. The list
+/// holds model values, which the typed layer already reduced to their keys.
+fn rewrite_embedded_relation_in_list(cx: &ExprContext, e: &mut stmt::ExprInList) {
+    if let Some(resolved) = resolve_embedded_relation(cx, &e.expr)
+        && resolved.tail.is_empty()
+    {
+        *e.expr = resolved.key_expr;
```
carllerche: [P2] Normalize composite foreign keys before lowering list membership

When the embedded relation has a composite key, this substitution produces a record of key projections, but `lower_expr_in_list` requires each record member to already be a bare column. Adding `Expr::in_list(variant().owner().parent().parent(), [&first, &second])` to the existing `nested_variant_composite_relation` setup (UUID + i64 primary key) panics at `lower.rs:1651` with `assertion failed: lhs.is_column()`, whereas the equivalent equality comparison succeeds. Normalize the tuple’s projected keys and decoding casts, together with the corresponding list values, before passing them to list-membership lowering.

crates/toasty-macros/src/model/expand/embedded_enum.rs
```diff
@@ -312,6 +320,230 @@ impl Expand<'_> {
...
+                                })
+                            });
+
+                            quote! {
+                                #vis fn #field_ident(
+                                    mut self,
+                                    value: impl #toasty::EmbeddedRelationValue<
+                                        #rel_ty,
+                                        Model = <#rel_ty as #toasty::RelationOneField>::Target,
+                                    >,
```
carllerche: [P2] Preserve type inference for complete relation literals

Rewriting a complete variant literal through this generic setter removes the contextual field type from expressions such as `human: Default::default()`. With `human: Deferred<Human>` declared on the variant, `create!(Object { id, owner: Owner::Human { id, human: Default::default() } })` compiles on the base commit but fails here with E0283: Rust cannot choose among the model, reference, and deferred implementations of `EmbeddedRelationValue`. Preserve the declared relation field’s type inference for ordinary complete literals while supporting the borrowed-model shorthand.

carllerche (commented):
Fresh review of dc901cf1acb2e03a5e1f687100c165511311f678, including an independent sub-agent pass. Three reproducible findings are attached.

Validation: all 808 existing SQLite tests, 12 focused core tests, and 4 required-literal/private-field tests pass. Added focused reproductions in isolated checkouts: optional-struct variant filtering and `Default::default()` relation literals both pass on base 5b4619b84 and fail on this head; composite-key embedded relation list membership also panics on this head.

crates/toasty-core/src/stmt/cx.rs
```diff
@@ -461,6 +461,24 @@ impl<'a, T: Resolve> ExprContext<'a, T> {
             }
             Expr::IsNull(_) => Type::Bool,
             Expr::IsVariant(_) => Type::Bool,
+            Expr::Variant(e) => {
```
carllerche: Are there unit tests for this?

carllerche:
Macro design feedback for `toasty-macros`: retain qualified variant-literal syntax and obtain construction builders through generated destination-field traversal in both `create!` and `update!`.

**[Variant construction and naming](https://github.com/tokio-rs/toasty/blob/79bdc3648f18594e7138936034b4b381962e1e37/crates/toasty-macros/src/variant_literal.rs#L80)**

Please obtain variant construction builders through the destination field's generated traversal, using the same approach as existing nested builders.

Keep `Owner::Human { human: &alice }` as the syntax that triggers variant construction. For example:

```rust
create!(Object {
    owner: Owner::Human { human: &alice },
})
```

should expand approximately to:

```rust
Object::create().owner(
    Object::fields().owner().create().human().human(&alice),
)
```

The destination field determines the enum type; the parsed final segment selects the variant. The first `.human()` selects the variant, and the second sets its relation field. This removes the need for a hidden `Owner::__expr_human()` constructor per variant.

Use existing naming conventions: `.create()` on the generated fields handle, an `OwnerCreate` variant-selection builder, and an `OwnerHumanCreate` payload builder. Derive variant method names through the existing `Name::from_ident` machinery used by field accessors, including its handling of acronyms and raw identifiers; avoid a separate snake-case implementation here. Store generated builder identifiers alongside the existing schema identifiers and reuse the naming helpers.

Use the same construction builder for both `create!` and replacement assignments in `update!`, with `IntoExpr` and replacement `Assign` support, including optional enum fields. A variant literal constructs a replacement value; existing anonymous update braces retain their partial-patch meaning. Keep the relation-key filling, order-independent precedence of loaded relation keys, and required-field validation in the shared builder.

Accepted syntax limitations: imported variants such as `Human { ... }`, including renamed imports, do not get enum-builder rewriting. Complete ordinary Rust values can still pass through unchanged. The enum qualifier may be ignored when selecting the destination builder, so `DifferentEnum::Human` may currently select the destination's `human()` builder. A compile-time qualifier/type check would be welcome if straightforward, but builder lookup should still use the destination field. No new variant DSL is needed.

**[Update expansion and evaluation order](https://github.com/tokio-rs/toasty/blob/79bdc3648f18594e7138936034b4b381962e1e37/crates/toasty-macros/src/update/expand.rs#L38)**

For update, extend the existing nested-builder expansion: hoist each variant field's value, then construct the variant builder through the destination fields after obtaining the update builder. The current `expand_create_setters` / `expand_has_many_list_item` path already follows this sequence.

For example, `update!(object { owner: Owner::Human { human: &alice } })` should expand approximately to:

```rust
match (&alice,) {
    (human,) => {
        let update = object.update();
        let fields = update.__macro_fields_root();
        update.owner(
            fields.owner().create().human().human(human),
        )
    }
}
```

Recognize and expand the variant where the destination field path is available, rather than rewriting the whole literal inside the generic `Hoist::hoist`. Share the variant parsing/builder expansion with `create!`, while letting update register leaf values with its existing hoister. For a replacement inside an embedded patch, traverse to that nested destination and pass the constructed value to the existing patch machinery.

Preserve the existing `match`-based temporary lifetimes, left-to-right value evaluation, and single target evaluation after the values. Test both instance and query updates, nested replacements, inference-dependent values, and borrowed temporaries.

There is a separate borrowing edge case: hoisting a reference into the update target itself retains that borrow across `.update()`. Existing nested creates already reject `title: user.name.as_str()` while accepting `title: user.name.clone()`; I verified both against the current nested-create implementation. The PR's whole-variant hoisting supports some additional cases by consuming those references earlier. Make that compatibility difference explicit in tests, but it does not prevent using the existing destination-field traversal for enum builders, and does not require resolving builders through the enum qualifier.

carllerche:
Follow-up on the construction entry point: please use a shared helper plus an internal trait in `codegen_support` to obtain the builder from the destination field handle.

Adding an inherent `create()` method to the enum's fields handle collides with the existing accessor for a data-carrying variant named `Create`. This valid declaration compiles on the base branch but fails after the refactor with E0592 (duplicate definitions named `create`):

```rust
#[derive(toasty::Embed)]
enum Action {
    Create { name: String },
    Delete,
}
```

There is already precedent for the helper/trait approach: [`scope_fields` and `create_in_scope`](https://github.com/tokio-rs/toasty/blob/56dcd310399ed23e862fbf0e3b6222d793cdf2dd/crates/toasty/src/codegen_support.rs#L45) accept a typed value, let Rust infer its concrete type, and dispatch through `Scope`. `ModelCodegen` in the same module also supplies internal constructors used by generated accessors.

Apply that pattern here, schematically:

```rust
// In codegen_support:
pub trait Create {
    type Create;
    fn create(&self) -> Self::Create;
}

pub fn create<F: Create>(fields: F) -> F::Create {
    <F as Create>::create(&fields)
}
```

The macro expansion then obtains its builder through the destination field:

```rust
toasty::codegen_support::create(Object::fields().owner())
    .human()
    .human(&alice)
```

Implement the trait for the relevant generated field handles and optional/list paths. Explicit trait dispatch avoids inherent-method lookup, so the existing accessor for `Action::Create` remains available. Remove the newly added inherent construction methods; no hidden per-variant constructor or new reserved variant name is needed.

Use the helper in both `create!` and `update!`. Keep update's leaf-value hoisting, the shared `OwnerCreate` / `OwnerHumanCreate` builders, and the qualifier check. This changes only how the construction builder is obtained. Add regression coverage for a `Create` variant, including its existing field accessor and construction through both macros.

carllerche (commented):
Two suggestions to reduce the additions in `crates/toasty/src/engine/lower/lift_in_subquery.rs`:

1. **Reuse the existing predicate-lifting path for embedded relations.** The new `lift_embedded_relation_comparison`, `lift_embedded_relation_like`, `lift_embedded_relation_is_variant`, and `lift_embedded_relation_predicate` functions repeat the existing predicate constructors and traversal logic. The difference is how the relation is located and how its source key is expressed; after resolution, both paths build a target-model predicate and a foreign-key subquery.

   Extend `lift_relation_path_predicate` to handle either an ordinary or embedded relation resolution, so the existing comparison, `LIKE`, and `IsVariant` wrappers serve both. Similarly, extend `lift_in_subquery` for embedded relation endpoints and reuse the existing traversal machinery for the remaining path. This should remove most of the second dispatch in `rewrite_embedded_relation` and the parallel `lift_embedded_relation_*` functions while preserving the existing `Has`/`Via` handling and paired-relation optimization. As a small first step, `embedded_fk_in_subquery` duplicates `lift_fk_in_subquery`, which already accepts an arbitrary LHS expression.

2. **Defer the subquery-merging optimization.** `merge_in_subqueries`, `mergeable_in_subqueries`, `simple_select`, and `returning_is_unique_key` add 108 lines, plus the visitor call and its explanation. Their correctness argument establishes that the two-subquery form has the same result when the selected key is unique:

   ```sql
   k IN (SELECT id FROM m WHERE a) AND k IN (SELECT id FROM m WHERE b)
   ```

   This appears to be a separate query optimization rather than a requirement for embedded relations. Remove it from this PR and run the relevant SQL and DynamoDB tests to confirm that both backends handle the unfused form. The expected tradeoff is additional query work. If a backend requires fusion, retain a narrower solution for that requirement; otherwise, deferring this removes roughly 117 lines from the PR.

These suggestions come from static review; the reductions have not been implemented or tested.

carllerche (commented):
Within-variant updates can stay in a follow-up PR, but this PR needs to reject them explicitly and correct the documentation before merging. Whole-value replacement through the setters and `update!` can remain supported.

The missing `Enum::fields()` method does not prevent callers from reaching the incomplete patch implementation. A path can enter a variant through an enclosing embedded struct:

```rust
#[derive(Debug, toasty::Embed)]
enum State {
    Active { count: i64 },
    Inactive,
}

#[derive(Debug, toasty::Embed)]
struct Details {
    unrelated: i64,
    state: State,
}

let patch = toasty::stmt::patch(
    Details::fields().state().active().count(),
    toasty::stmt::increment(),
);
```

`stmt::patch` currently keeps only `path.untyped.projection`, discarding the variant root and its parent path. In this example, it keeps `count`'s variant-local index, `0`. Applying this patch to a model's `details` field therefore targets `details.unrelated`, not `details.state.Active.count`.

I reproduced the assignment construction against head `90d50758c8b3538978d278659cf37f096689563a`: applying the patch at model-field index `2` produces `Projection(2, 0): Add(I64(1))`. This checks the generated assignment; it is not a database execution test. The relevant code is [`stmt::patch`](https://github.com/tokio-rs/toasty/blob/90d50758c8b3538978d278659cf37f096689563a/crates/toasty/src/stmt/assignment.rs#L370). A check only in assignment lowering cannot catch this reliably because the truncated projection can name a valid unrelated field.

Please make the following changes in this PR:

1. **Reject patches that traverse an enum variant before discarding the path root.** Cover paths that enter a variant through enclosing structs or other variants. Prefer recording a deferred construction error and returning `unsupported_feature` before planning or executing the statement. `patch()` is currently infallible, so this needs error propagation rather than an immediate `Result` from the existing API. Reject the entire statement before any writes, including other assignments in the same statement.
2. **Correct the documentation.** The [within-variant patch paragraph](https://github.com/tokio-rs/toasty/blob/90d50758c8b3538978d278659cf37f096689563a/docs/dev/design/polymorphic-relations.md#L325) incorrectly claims existing variant-gated, SQL-only support. State that within-variant patches, including increments and relation-key changes, are unsupported on every backend. Replace the nonworking enum examples in `stmt::patch` rustdoc with supported embedded-struct examples.
3. **Add regression coverage for the boundary.** Test rejection of both setting and incrementing variant fields, including the enclosing-struct example where the truncated index names a valid unrelated field. Verify that rejection leaves all fields unchanged and performs no writes. Keep positive coverage for ordinary embedded-struct patches, whole-enum replacement, and replacing an enum field nested inside a struct with `stmt::patch`.

The rejection should apply to traversing into a variant, not to assigning an enum value. Replacing an enum field inside a struct must continue to work. Variant-aware assignments, matching/nonmatching-variant semantics, backend support, and macro syntax can all remain in the follow-up; this PR does not need a new capability flag or a partial implementation of those features.

carllerche (commented):
Reviewed head `c1bfb39a`. The strongest remaining opportunities to reduce code and the PR diff are:

1. **Construct variant builders directly from the enum named in the literal.** `VariantLiteral::expand` in `crates/toasty-macros/src/variant_literal.rs` already knows the enum qualifier in `Owner::Human { ... }`. It currently obtains a builder through the destination field, then checks the qualifier with `variant_of`. It could instead generate:

   ```rust
   <Owner as toasty::codegen_support::EmbedCreate>::create()
       .human()
       .human(&alice)
   ```

   This removes the new `codegen_support::Create` trait, `create` helper, `variant_of`, generated field-handle implementations, optional-path implementations, and destination plumbing in the create/update macros. Estimated reduction: over 100 lines, including the associated explanatory comments.

   Keep the existing builders, monomorphic setters, and update expression hoisting. A wrong-enum literal still fails at the destination setter, so its compiler diagnostic snapshot would change. Explicit trait invocation also handles the `Create` variant naming collision.

2. **Use the existing lowering hooks for relation-to-key substitution.** The new endpoint rewrite functions in `engine/lower/lift_in_subquery.rs` repeat an equality-operator check and traversal of both operands. The binary rewrite also returns a boolean that its caller no longer uses.

   `LowerStatement` already has hooks that perform these checks and rewrite operands before lowering their children. Add one small embedded-relation-to-key helper there, shared by `rewrite_eq_operand` and the list-membership hook. This removes both new expression-specific wrappers and their prepass dispatch, for an estimated 25–40 lines of reduction. Keep relation traversal and conjunction lifting in the prepass; only move endpoint substitution. Preserve the existing model-membership restrictions and substitution on both comparison operands.

3. **Share schema traversal between verification and embedded relation resolution.** `Level` in `engine/lower/embedded_relation.rs` and `PathTarget` resolution in `engine/verify.rs` independently implement variant selection and field lookup. Share those operations: verification walks to the endpoint; relation resolution stops at the first `belongs_to` and builds sibling-key projections.

   This should remove duplicate logic if the shared helper stays small; the net savings need a prototype. Preserve document handling, scope validation, and the distinction between flattened schema field IDs and variant-local expression positions. Reverting verification to `Schema::resolve` would lose that distinction.

4. **Two straightforward deletions, totaling about 22 lines.** In `expand_enum_create_builders`, reuse `relation_key_fill(&fields)` instead of constructing another `key_sources` map. The diagnostic match becomes `Some((rel_ident, _))`. In `simplify_expr_project`, remove the new `Variant` early return: the remaining branches only handle `Value`, `Record`, and `Match`, so `Variant` already falls through to `None`. The latter removes that file's entire diff.

One correction to the earlier review: retain the narrower conjunction grouping. The revised implementation documents the DynamoDB planner's one-substatement-input limitation, so removing grouping outright would require addressing that constraint.

These are static-review recommendations; the reductions have not been implemented or tested.

crates/toasty-core/src/stmt/path.rs
```diff
@@ -106,14 +110,30 @@ impl Path {
...
+    /// `other`'s root model is dropped: its steps continue from wherever
+    /// `self` ends. A variant selection in `other` is preserved — the result
+    /// selects the same variant at the same point of the traversal — so a
+    /// variant-rooted path can be chained onto a path reaching its enum from
+    /// another model.
     pub fn chain(&mut self, other: &Self) {
+        if let PathRoot::Variant { parent, variant_id } = &other.root {
+            self.chain(parent);
+            *self = Self::from_variant(self.clone(), *variant_id);
+        }
```
carllerche: Are there tests for this new code path? If not, add some.

PR #1182, opened 2026-08-11
feat(engine): add conditional execution to the exec program

carllerche:
## Summary

Fixes eager (non-deferred) `belongs_to` loading on the insert paths, and adds the exec program's first control-flow construct — which the upsert fix needs to be correct without a per-case workaround.

- **`INSERT` model returning** — creating a model with an eager `belongs_to` panicked in the planner, in every shape (single `create!`, batch, nested). Relation load subqueries are now built per VALUES row, after relation planning resolves the FK fields, so each subquery binds to its row index. Filter args on a query child of an insert parent rewrite to row-projected inputs (`rewrite_stmt_query_arg_dependencies`), mirroring the existing assignment rewrite. Nested creates insert children before the parent, making the subquery's ordering edge cyclic at statement granularity while acyclic at operation granularity; lowering tracks the enclosing-insert stack and records the edges as `DepKind::Effect` deps anchored on the target's data-loading node.
- **`DO NOTHING` upsert returning** — the per-row returning pass skipped the projection preserved for `.or_ignore()` upserts, so a successful insert returned the relation slot as `Null` and decoding failed. Insert projection returning now accepts sub-statement args, evaluated per database row via `map` with inputs one scope out.
- **Exec control flow** — on conflict the upsert's `RETURNING` produces zero rows, so the relation-load `SELECT` has nothing to correlate on. A structured `Action::If` now skips a run of pure actions when a guard condition fails; a plan-time annotation pass decides what is guarded (map-body-only consumption in an `Eval` over a node's rows, plus the guarded-consumer closure). Variable refcounts become exact along the way: `num_uses` counts per-load value edges plus the exit use, ordering-only deps no longer pin slots, zero-use outputs are not stored, and a debug teardown assertion checks every slot drains. The conflict path issues one statement instead of two, and back-refs read the insert's `RETURNING` rows — safe because the reader only runs when a row exists.
- **Planner cleanup on top of the new machinery** — ordering edges between HIR statements become `IndexMap<StmtId, DepKind>`: a `Statement` edge anchors on the target's output node, an `Effect` edge on its data-loading node, so the anchor is declared where the edge is created instead of decided by planning-order timing. `mir::Store` gains reserved slots filled with an `Alias` pass-through, replacing the `deferred_node_deps` patch-later list. The pre-existing `Guard` action is folded into `If` via a `Cond::Expr` variant (its one use, the KV update pre-filter, becomes a guarded `Alias` of the key list) — net deletion, and it collapses the last `input_loads()`/`inputs()` special case. `ExecStatement` mutation classification now uses `effect::classify` (its first consumer) instead of a `ConditionalOutput` proxy.

The design doc for the control-flow work ([rendered](https://github.com/tokio-rs/toasty/blob/fix-create-eager-belongs-to/docs/dev/design/exec-control-flow.md)) rides on this branch for review context only and is removed before merge.

## Type of change

- [x] Other (please explain): two bug fixes plus the engine-internal control-flow mechanism they motivated. No user-facing API or driver contract changes; some plans issue fewer driver operations.

## Checklist

- [x] `cargo test` passes

## Notes for reviewers

- The regression test `eager_belongs_to_upsert_or_ignore_conflict_issues_no_load` asserts via the driver op log that the conflict path issues no read. With the guard annotation disabled it fails (the unguarded `SELECT` substitutes from the empty `RETURNING` and panics), so it pins the behavior.
- The debug teardown assertion (`VarStore::assert_empty`) runs in every suite test, which is what validates the exact-refcount change across the plan space.
- `toasty-core` gains one public item: `stmt::visit::walk_expr_scoped`, a read-only mirror of the existing `walk_expr_scoped_mut`, used by the guard analysis.
- Control-flow commit verified on SQLite (790/790), DynamoDB against a local instance, the full workspace, clippy, fmt, and the WASM check; PostgreSQL/MySQL services were not available in that environment. The two fix commits were previously verified against all four drivers (SQLite 789, PostgreSQL 818, MySQL 779, DynamoDB 620).
- The `Guard`→`If` fold preserves semantics: `Cond::Expr` inputs are counted loads on both arms, `Expr` guards never share an `If` block, and the annotation pass neither overwrites nor propagates them (their else-arm placeholder is an observed value, not proven unobservable). The folded path is exercised 19 times across the DynamoDB suite.
- Known residual limitation (pre-existing, unchanged): an inline nested create through the belongs_to side (`Post::create().user(User::create())`) with a DB-generated auto-increment key on MySQL still cannot thread the non-constant FK through `RETURNING`.


carllerche:
@codex review this

carllerche:
@codex review this.

carllerche:
@codex Review this again

carllerche:
@codex review this again

PR #1195, opened 2026-08-19
fix: get Connect max_connection from driver

### LINKED ISSUE #1125: Updating multiple records in future::join_all failing
### What you did

```
futures::future::join_all(users.iter_mut().map(async |user| -> toasty::Result<()> {
        println!("updating {:?}", user);
        toasty::update!(user {
            todos: [{ title: "Take a walk" }]
        })
        .exec(&mut db.clone())
        .await?;
        Ok(())
    }))
    .await;
```

I was playing with toasty a while back and I was testing updating multiple records using join_all and it worked.

after updating toasty its now failing with `Error: no such table: todos: Error code 1: SQL logic error`

### What you expected

_No response_

### What actually happened

_No response_

### Minimal reproducer

```rust

```

### Database driver

_No response_

### Toasty version

_No response_

### Additional context

_No response_

Gitigi:
## Summary

Have Connect implement max_connection that is retrieved from the underlying driver
fixes #1125 having multiple connection to sqlite-memory failing since sqlite-memory does not support connection pool

## Type of change

## Checklist

- [x] `cargo test` passes

### DIFF
--- a/crates/toasty/src/db/connect.rs
+++ b/crates/toasty/src/db/connect.rs
@@ -133,6 +133,10 @@ impl Driver for Connect {
         self.driver.connect(cx).await
     }
 
+    fn max_connections(&self) -> Option<usize> {
+        self.driver.max_connections()
+    }
+
     fn generate_migration(&self, schema_diff: &diff::Schema<'_>) -> Migration {
         self.driver.generate_migration(schema_diff)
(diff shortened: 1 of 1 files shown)


joeydewaal:
Thanks for looking into this! LGTM

PR #1207, opened 2026-09-01
feat: Path field metadata

y-l-g:
## Summary

New accessors on `Path<M, T>`:

- `field_name()`
- `is_nullable()`
- `is_unique()`

## Type of change

- [x] Other : accessors

## Checklist

- [x] `cargo test` passes
- [x] Touches a public API → a design doc under `docs/dev/design/`
      describes the behavior — read-only views over the existing app
      schema; no new behavior or driver contract


joeydewaal:
Thanks for the pr. Out of curiosity, what's your use case for this? `is_nullable` and `is_unique` make sense I think but why only expose getters for app level and not db level names for fields?

y-l-g:
Hi! I'm building a generic admin panel (inspired by Filament for Laravel). I want field_name() for column labels and form input keys, and is_nullable() / is_unique() for form validation. I don't need the db level name as I don't write raw SQL, the app-level name is fine for these cases.

joeydewaal:
Ah that makes sense. Ping me when this is ready for a review!

crates/toasty/src/stmt/path.rs
```diff
@@ -797,6 +797,260 @@ impl<T, U> From<Path<T, U>> for stmt::Path {
...
+                == Some(shared)
+        })
+    }
+
+    /// Collects the model set rooted at `T` so field lookups can walk into
+    /// embedded models. Built per call: there is no global registry, and
+    /// `T::register` rebuilds the schema of every reachable model, so the
+    /// metadata accessors built on this are for one-off use, not per-row
+    /// loops.
+    fn registered_models() -> app::ModelSet {
```
joeydewaal, line 909: nit: maybe change the name of this function. `registered_models` implies this has something to do with the registered models in `Db` but this basically builds the whole tree of reachable models from `T`.

crates/toasty/src/stmt/path.rs
```diff
@@ -797,6 +797,260 @@ impl<T, U> From<Path<T, U>> for stmt::Path {
...
+    /// # #[derive(Debug, toasty::Model)]
+    /// # struct User {
+    /// #     #[key]
+    /// #     id: i64,
+    /// #     name: String,
+    /// # }
+    /// let name = User::fields().name().field_name();
+    /// assert_eq!(name, "name");
+    /// ```
+    pub fn field_name(&self) -> String {
```
joeydewaal, line 827: IMO this should return an `Option<String>` wdyt?

crates/toasty/src/stmt.rs
```diff
@@ -79,6 +79,11 @@ pub use upsert::Upsert;
 pub use toasty_core::stmt::BigDecimal;
 #[cfg(feature = "rust_decimal")]
 pub use toasty_core::stmt::Decimal;
+/// The untyped path underlying the typed [`Path`], produced by its `From`
+/// conversion.
+pub use toasty_core::stmt::Path as CorePath;
+/// An ordered sequence of values representing a record (row).
+pub use toasty_core::stmt::ValueRecord;
```
joeydewaal, line 88: Do we need to make these public? We should probably try to avoid that.

joeydewaal:
@codex review

crates/toasty/src/stmt/path.rs
```diff
@@ -797,6 +797,260 @@ impl<T, U> From<Path<T, U>> for stmt::Path {
...
+    }
+
+    /// The embedded model a projection step traverses into: a column-expanded
+    /// embed (`Embedded`) or a `#[document]` field (`Primitive(Model)`), whose
+    /// accessor is chainable like a column-expanded embed.
+    fn embedded_target(field: &app::Field) -> ModelId {
+        match &field.ty {
+            app::FieldTy::Embedded(embedded) => embedded.target,
+            app::FieldTy::Primitive(primitive) if let stmt::Type::Model(id) = &primitive.ty => *id,
+            _ => panic!("cannot project through non-embedded field"),
```
joeydewaal, line 971: We want to support non-embedded fields right? Something like `User::fields().todos().id().field_name()` panics now.

PR #1217, opened 2026-09-11
feat: reference embedded relations with model values

carllerche:
Implement step 2 of the polymorphic-relations design: a relation stored in an embedded type can now be compared to and set from a model value.

Filters resolve a projection into an embed to the belongs_to it lands on or passes through, using the is_variant gates the typed layer fuses next to variant-rooted comparisons (enum record slots are variant-local and ambiguous without them). A comparison at the relation substitutes the sibling key slot(s); traversal into the target model and in_query lift to a foreign-key IN subquery. An enum decode Match on the LHS of a membership test is eliminated eagerly at lower time into variant-gated terms, since on non-SQL backends the subquery detaches into its own statement before post-lower simplification could reach it. The verify pass gains a record-slot-space resolver for expression projections; Schema::resolve is path-space and only accepted step-1 shapes by accident.

On the write side, generated IntoExpr bodies fill key slots from a loaded relation value by reading the referenced field off the parent model, and create!/update! rewrite variant literals into derive- generated expression builders, so the parent can be passed by reference with the key omitted: Owner::Human { human: &alice }. Non-relation builder setters are monomorphic in the declared field type so values keep type-checking exactly as in a plain struct literal.


carllerche:
Was just curious

PR #1218, opened 2026-09-12
fix(toasty-core): resolve enum fields by variant-local index

y-l-g:
## Summary

`Schema::resolve` reads the field step after an enum variant as a global index into `EmbeddedEnum::fields`. The rest of the codebase uses a local index within the variant. This PR switches `resolve` to `variant_fields(variant).nth(local)`.

```rust
struct User {
    id: i64, // 0
    name: String, // 1
    status: Status, // 2
    contact: ContactInfo, // 3
}
enum ContactInfo {
    Email { address: String }, // variant 0
    Phone { country_code: String, number: String }, // variant 1
}

// before
assert_eq!(
    s.resolve_field(root, &Projection::from([3, 0, 1])).unwrap().name.app.as_deref(),
    Some("country_code")
);

// after
assert!(s.resolve_field(root, &Projection::from([3, 0, 1])).is_none());

```
## Type of change

- [x] Small / obvious change: bug fix, docs, internal cleanup, or test

## Checklist

- [x] `cargo test` passes

### DIFF
--- a/crates/toasty-core/src/schema/app/schema.rs
+++ b/crates/toasty-core/src/schema/app/schema.rs
@@ -204,12 +204,14 @@ impl Schema {
                             current_field = s.fields.get(*step)?;
                         }
                         Model::EmbeddedEnum(e) => {
-                            let variant = e.variants.get(*step)?;
+                            let variant_index = *step;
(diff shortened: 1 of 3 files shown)


joeydewaal:
Thanks!

PR #1227, opened 2026-09-15
feat: add explicit support for MariaDB

joeydewaal:
## Summary
Since we support mysql users can connect and use MariaDB since they have the same transport protocol. MariaDB however supports some features that mysql doesn't, like native support for UUID and returning from an insert statement. By adding explicit MariaDB support in toasty users and the engine can use those native features.

## Type of change

- [ ] Other (please explain):

## Checklist

- [x] `cargo test` passes

## Notes for reviewers
N/A


carllerche:
Thanks for this. I'm going to work a bit on this PR. I think a better way might be to just introduce a new MariaDB driver since they are forks.

PR #1238, opened 2026-09-20
fix: derive PartialEq, Eq, and Hash for Deferred<T>

### LINKED ISSUE #1236: Deferred&lt;T&gt; doesn't derive PartialEq/Eq, forcing manual impls on any model with a relation field
### Summary

`Deferred<T>` only derives `Clone`:

```rust
// toasty-0.10.0/src/schema/deferred.rs
#[derive(Clone)]
pub struct Deferred<T> {
    value: Option<Box<T>>,
}
```

Since `#[belongs_to]` and `#[has_many]` fields are wrapped in `Deferred<T>`, adding a relation to a model makes `#[derive(PartialEq, Eq)]` on that model stop compiling. The only workaround is a hand-written `impl PartialEq` that lists every persisted field and excludes the relation fields, e.g.:

```rust
#[derive(Clone, Debug, toasty::Model)]
pub struct Campaign {
    #[key] #[auto]
    pub id: CampaignId,
    pub name: String,
    // ...several more scalar fields...

    #[has_many]
    pub capabilities: toasty::Deferred<Vec<CampaignCapability>>,
}

// `toasty::Deferred` does not implement `PartialEq`/`Eq`, so equality is defined
// manually over the persisted fields; the `capabilities` relation is excluded.
impl PartialEq for Campaign {
    fn eq(&self, other: &Self) -> bool {
        self.id == other.id && self.name == other.name /* ...and so on... */
    }
}
```

This scales badly: every new persisted field has to be added to the manual impl by hand, and nothing enforces that — the compiler happily accepts an impl that silently ignores a newly added field.

### Suggested fix

Derive `PartialEq`/`Eq` (and probably `Hash`) on `Deferred<T>`, bounded on `T`:

```rust
#[derive(Clone)]
pub struct Deferred<T> {
    value: Option<Box<T>>,
}

impl<T: PartialEq> PartialEq for Deferred<T> {
    fn eq(&self, other: &Self) -> bool {
        self.value == other.value
    }
}
impl<T: Eq> Eq for Deferred<T> {}
```

This would let `#[derive(PartialEq, Eq)]` work again on models with relation fields, comparing loaded values when both sides are loaded.

### Toasty version

0.10.0

shuvroroy:
## Summary

Allow models with `Deferred<T>` relation fields to derive `PartialEq`, `Eq`, and `Hash` by deriving these traits on `Deferred<T>`, bounded on `T`. Equality and hashing include the load state and loaded value; an unloaded field differs from a loaded `None`.

Add five regression tests covering load states, nullable values, types that implement only `PartialEq`, hashing, and models with `has_many` and `belongs_to` relations.

Closes #1236.

## Type of change

## Checklist

- [x] `cargo test` passes
- [ ] Touches a public API → a design doc under `docs/dev/design/` describes the behavior no separate design doc for this small trait-derive fix; the behavior is documented in `Deferred`'s rustdoc.

## Notes for reviewers

The new tests fail to compile before the fix and pass afterward. The full `cargo test` run passes: 2,859 passed, 0 failed, 310 ignored.


carllerche (approved):
Thanks!

PR #1239, opened 2026-09-20
fix: honor non-primary references in relation comparisons

### LINKED ISSUE #1219: Comparing a belongs_to relation to a model value ignores non-primary `references`
### What you did

Filtered a model by comparing a `belongs_to` relation to a model value. The relation uses a non-primary `references` field.

```rust
Gadget::filter(Gadget::fields().bot().eq(&bot))
```

`Gadget.bot` is declared with `#[belongs_to(key = bot_serial, references = serial)]`. `Bot.serial` is a unique `String`. `Bot.id` is the primary key and is a `uuid::Uuid`.

### What you expected

One row: the gadget whose `bot_serial` column equals `bot.serial`.

### What actually happened

The query compares the `bot_serial` column to `bot.id`, not to `bot.serial`. On SQLite the value encoder panics:

```
thread 'comment2_root_non_pk_reference_eq_model' panicked at crates/toasty-driver-sqlite/src/value.rs:104:18:
not yet implemented: value = Uuid(
    01a095ed-96bd-73c1-85fd-515e7c304f97,
)
```

When the key and the primary key have the same type, the query does not panic. It compares the wrong values and returns no rows.

### Minimal reproducer

```rust
struct Bot {
    #[key]
    #[auto]
    id: uuid::Uuid,
    #[unique]
    serial: String,
    name: String,
}

struct Gadget {
    #[key]
    #[auto]
    id: uuid::Uuid,
    #[index]
    bot_serial: String,
    #[belongs_to(key = bot_serial, references = serial)]
    bot: toasty::Deferred<Bot>,
}

#[tokio::test]
async fn root_non_pk_reference_eq_model() {
    let mut builder = toasty::Db::builder();
    builder.models(toasty::models!(Bot, Gadget));
    let mut db = builder
        .build(toasty_driver_sqlite::Sqlite::in_memory())
        .await
        .unwrap();
    db.push_schema().await.unwrap();

    let bot = toasty::create!(Bot { serial: "B-1", name: "Marvin" })
        .exec(&mut db)
        .await
        .unwrap();
    let g = toasty::create!(Gadget { bot_serial: "B-1" })
        .exec(&mut db)
        .await
        .unwrap();

    let found: Vec<Gadget> = Gadget::filter(Gadget::fields().bot().eq(&bot))
        .exec(&mut db)
        .await
        .unwrap();
    assert_eq!(found.len(), 1);
    assert_eq!(found[0].id, g.id);
}
```

### Database driver

SQLite (the panic). The wrong comparison is not driver-specific.

### Toasty version

commit 13ae82bc0 (branch `feat-polymorphic-relations`, also present on `main` at e3f26668a)

### Additional context

The same comparison fails in the same way for relations inside embedded types, added in #1172. `Expr::in_list` with a relation path has the same problem.

shuvroroy:
## Summary

Fix `belongs_to` comparisons that use the target's primary key when `references` names a different unique field. Preserve key values until lowering selects the referenced fields for `eq`, `ne`, and `in_list`, including relations inside embedded structs and enums.

Closes #1219.

## Type of change

## Checklist

- [x] `cargo test` passes

## Notes for reviewers

Four integration tests reproduce the panic with differing key types and incorrect results with matching key types. They also cover embedded relations, primary-key identity, owned model values, and unloaded deferred fields. All four fail on the original code and pass with the fix.

Validated with `cargo fmt`, `cargo clippy --offline`, and `cargo test --offline`.


carllerche:
This makes sense. Instead of preserving any indexed field, I would preserve fields that are known to be used as FKs in relations. The schema for the model can track that data and it can be populated when the model relation data is populated when building the schema

shuvroroy:
@carllerche Thanks, updated accordingly. The model schema now tracks fields referenced by relation FKs, populated when relations are linked during schema building. Simplification uses that metadata to preserve primary keys and referenced fields. I also added tests covering unindexed references and verifying that unrelated indexed fields are discarded.

PR #1255, opened 2026-09-23
docs: preserve database value comparisons in the Option contract

carllerche:
## Summary

[Rendered](https://github.com/tokio-rs/toasty/blob/docs-option-payload-semantics/docs/dev/design/option-semantics.md)

The Option contract currently requires Rust equality for present values,
which changes MySQL string collation and PostgreSQL NaN behavior. Restrict
Rust rules to presence: `None == None`, `None != Some(_)`, and two `Some`
values compare using the database's value equality.

Keep native value ordering for pagination, including PostgreSQL NaN and
database-equal sort values with a unique tie-breaker. Move field, row, and
loading presence into Behavior and explain them with the nickname example.
Update the roadmap to distinguish presence tests from backend value tests.

Follow-up to #1254 addressing both review findings. Related: #188.

## Type of change

## Checklist

- [x] `cargo fmt` is clean
- [x] Focused existing integration tests pass
- [x] The design describes application presence and driver-observable value comparisons

## Notes for reviewers

Audit findings on `main`:

- Field comparisons emit native `=` and `<>`. The document-field equality
  test deliberately matches the plain column's backend collation.
- Constant folding and client evaluation use Rust value comparisons.
  In-memory relation matching uses local hash and sort comparisons. The
  roadmap identifies these paths for backend agreement checks.
- Native collection membership keeps its own comparison contract; for
  example, JSON string membership stays case-sensitive on MySQL/MariaDB.
- Cursor predicates use native value comparisons and add primary-key
  tie-breakers when needed. The contract now preserves that value ordering
  while defining Option absence placement.

Validation: 12 existing integration cases passed across SQLite, PostgreSQL,
MySQL, and MariaDB (`struct_embed_filter_matches_column_case_sensitivity`,
`query_in_list::in_list_with_null`, and `vec_string_contains_filter`).
Direct database probes confirmed case/trailing-space equality under
`utf8mb4_unicode_ci`, case distinction under `utf8mb4_bin`, and PostgreSQL's
NaN equality, membership, and ordering. Formatting, whitespace, local links,
and design section checks pass. This PR changes documentation only.


carllerche:
@codex review again

PR #1259, opened 2026-09-23
fix(engine): bind subquery inputs in key-value filters

carllerche:
## Summary

Filters that combine subquery results can panic or bind the wrong input on DynamoDB. Carry all dependencies through scans and partition queries, and retain subquery inputs for filters applied after a key lookup.

Keep query and mutation argument positions distinct. Bind mutation key and row filters through one shared input mapping so assignment dependencies cannot shift their values. Add integration coverage for combined and partitioned relation queries and a planner regression for reordered mutation dependencies.

## Type of change

## Checklist

- [x] `cargo test` passes for the engine unit tests and full SQLite and DynamoDB driver suites

## Notes for reviewers

Merge order: #1258 → #1259 → #1260 → #1261 → #1247. Each PR targets the preceding branch; rebase and retarget the next PR onto `main` after a squash merge.

Validation: 394 engine unit tests, 848 SQLite tests, and 674 DynamoDB tests pass. The driver suites use `cargo test -p tests --features dynamodb --test <driver>`; DynamoDB runs with `--test-threads=1`.


carllerche:
Addressed the [argument-scoping review](https://github.com/tokio-rs/toasty/pull/1259#pullrequestreview-5298155772) in 4166882aa. Dependency rewriting now tracks `Map`/`Let` scopes, preserves local bindings, and retains nesting when replacing both subquery arguments and parent-row references. The new regression reproduced the panic and now verifies the evaluated predicate with reordered dependencies. All 396 engine unit tests, 849 SQLite tests, and 676 DynamoDB tests pass.

